PR #290, opened 2013-09-01
MRG: thinking of an interface for HRF convolution

matthew-brett:
I'm trying to work out a good interface to express the fact that generally,
when we convolve things with HRFs, we want the same interval for the kernel,
and the same dt.

Maybe we could have an HRF object, that has a 'convolve' method, maybe an
'evaluate' method (to evaluation the real function at given time values, and a
`symfunc` property for the symbolic function.

I haven't really thought this last part through.

### DIFF
--- a/nipy/modalities/fmri/tests/test_utils.py
+++ b/nipy/modalities/fmri/tests/test_utils.py
@@ -23,11 +23,13 @@ from ..utils import (
     interp,
     linear_interp,
     step_function,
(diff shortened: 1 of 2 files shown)


matthew-brett:
Any comments here?  I think this code is an improvement, and we can merge this and think of the final interface for the convolver class later.

matthew-brett:
Last call for comments.

PR #301, opened 2013-10-31
Contrast

jonathan-taylor:
Tried adding a "show_contrast" method to DesignMatrix.

Also added cmap arguments to both show* methods.


matthew-brett:
OK - I will rebase this one, add some tests.

PR #309, opened 2014-02-26
Resample

jbpoline:
Pass some ndimage keyword arguments to the resampling routines. Useful if you need default value to something else than 0 for instance. Not all cases are covered.


matthew-brett:
Closing this one in favor of https://github.com/nipy/nipy/pull/359 (same PR with refactoring on top).

Issue #310, opened 2014-03-06
Paradigm class: unexpected behavior if con_ids are strings


mekman:
The following illustrates an unexpected effect when using the Paradigm class with condition names that are strings. The problem is in this line:
https://github.com/nipy/nipy/blob/master/nipy/modalities/fmri/experimental_paradigm.py#L51

    from nipy.modalities.fmri.experimental_paradigm import BlockParadigm
    onsets = np.array([0, 6])
    durations = np.array([1, 1])
    conditions = np.array(['001', '002'])

    paradigm = BlockParadigm(conditions, onsets, durations, amplitude=None)
    print paradigm.con_id

This yields the following:

    array(['c1', 'c2'], 
          dtype='|S2')
      
 this works fine  
      
    conditions = np.array(['C001', 'C002'])
    paradigm = BlockParadigm(conditions, onsets, durations, amplitude=None)
    print paradigm.con_id

    array(['C001', 'C002'], 
          dtype='|S4')

The docstring suggests that ``con_id`` should be a string. Therefore I think one would expect the latter result for both cases. 

Best,
 Matthias

matthew-brett:
Closed by #316

PR #316, opened 2014-04-08
Clean paradigm

bthirion:
Addresses issue https://github.com/nipy/nipy/issues/310

### DIFF
--- a/nipy/modalities/fmri/experimental_paradigm.py
+++ b/nipy/modalities/fmri/experimental_paradigm.py
@@ -22,6 +22,7 @@ Author: Bertrand Thirion, 2009-2011
 import numpy as np
 
 from ...utils.compat3 import open4csv
+import warnings
 
 ##########################################################
 # Paradigm handling
@@ -51,11 +52,16 @@ class Paradigm(object):
             try:
                 # this is only for backward compatibility:
                 #if con_id were integers, they become a string
+                warnings.warn(
+                    'The behavior will change,' +
+                    'prepending by a c will disappear in the future')
                 self.con_id = np.array(['c' + str(int(float(c)))
-                                        for c in con_id])
+                                         for c in con_id])
             except:
                 self.con_id = np.ravel(np.array(con_id)).astype('str')
-
+            # the following line should replace the try.. except statement
+            # self.con_id = np.ravel(np.array(con_id)).astype('str')
+            
         if onset is not None:
             if len(onset) != self.n_events:
                 raise ValueError(
--- a/nipy/modalities/fmri/experimental_paradigm.py
+++ b/nipy/modalities/fmri/experimental_paradigm.py
@@ -49,19 +49,7 @@ class Paradigm(object):
         self.amplitude = amplitude
         if con_id is not None:
             self.n_events = len(con_id)
-            try:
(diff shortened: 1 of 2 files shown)


nipy/modalities/fmri/experimental_paradigm.py
```diff
@@ -51,11 +52,16 @@ def __init__(self, con_id=None, onset=None, amplitude=None):
             try:
                 # this is only for backward compatibility:
                 #if con_id were integers, they become a string
+                warnings.warn(
+                    'The behavior will change,' +
+                    'prepending by a c will disappear in the future')
                 self.con_id = np.array(['c' + str(int(float(c)))
-                                        for c in con_id])
+                                         for c in con_id])
             except:
```
matthew-brett: What error are you trying to catch with the `except`?   Should try and avoid bare `except` if at all possible.

nipy/modalities/fmri/tests/test_paradigm.py
```diff
@@ -77,6 +77,15 @@ def test_read_paradigm():
     assert (read_paradigm.onset == paradigm.onset).all()
 
 
+def test_paradigm_with_int_condition_ids():
+    paradigm1 = basic_paradigm()
+    conditions = [0, 0, 0, 1, 1, 1, 2, 2, 2]
+    paradigm2 =  EventRelatedParadigm(conditions, paradigm1.onset)
+    #assert (paradigm2.con_id == np.array(conditions).astype('str')).all()
```
matthew-brett: Remove comment?  Or expand to say why comment remains?

bthirion: You're right on both points. I have edited the code accordingly (after a rebase).
Best,

nipy/modalities/fmri/experimental_paradigm.py
```diff
@@ -48,13 +48,7 @@ def __init__(self, con_id=None, onset=None, amplitude=None):
         self.amplitude = amplitude
         if con_id is not None:
             self.n_events = len(con_id)
-            try:
```
matthew-brett: I'm still worried by backwards compatibility - I think previous behavior was:

```
import nipy.modalities.fmri.experimental_paradigm as ep
In [2]: ep.Paradigm([1, 2, 3]).con_id
Out[2]: 
array(['c1', 'c2', 'c3'], 
      dtype='|S2')
```

where behavior on this branch is:

```
In [3]: ep.Paradigm([1, 2, 3]).con_id
Out[3]: 
array(['1', '2', '3'], 
      dtype='|S1')
```

I can see that it's a good idea not to convert strings to ints then back to strings, but maybe preserve `c` prefix if condition names are integers?

bthirion: I think that in general it is a bad idea to decide that the conditions cannot be named '1', '2' etc. and to change this for the user. Now, if you think that backward compatibility is more important, I can revert that one.

matthew-brett: It's OK, I can see arguments both ways...

matthew-brett:
Thanks - sorry to be slow in merging.

PR #317, opened 2014-04-28
Reggy affine

arokem:
I think I might have misplaced the translations for the affine in the previous PR. Just for my curiosity: would anyone care to enlighten me as to why these things have 5x5 affines?

### DIFF
--- a/nipy/algorithms/registration/scripting.py
+++ b/nipy/algorithms/registration/scripting.py
@@ -164,7 +164,7 @@ def space_time_realign(input, tr, slice_order='descending', slice_dim=2,
             # We use the top 4 by 4 as the affine for the new file we will
             # create:
             new_aff = new_im.affine[:4, :4]
-            new_aff[:, 3] = im.affine[:4, 4]
+            new_aff[:, 3] = new_im.affine[:4, 4]
             new_ni = nib.Nifti1Image(new_data, new_aff)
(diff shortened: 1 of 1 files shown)


matthew-brett:
Sorry for the crazy long time to get to this one.

I think the right way to solve this is to use `nipy.save_image` - see https://github.com/matthew-brett/nipy/commit/7c696a5d0dab634deccc543da687c80e55e46ff8

Does that look right to you?   Can you play with the output to check?

matthew-brett:
@arokem - do you have time to look at this one?

arokem:
Not today

arokem:
More precisely: this looks about right to me, but I will not have time to
experiment with this before October 12th. In particular, I don't quite
remember what data analysis triggered this fix, so I will have to dig up
the data I was probably using this with and investigate.

matthew-brett:
I think we're going to release before then, do you think I should release with this fix?

arokem:
Definitely with one of the fixes, probably yours - I probably would've gone
with that, if I was more nipy-savvy.

matthew-brett:
Ok - I hope my fix is right then.

matthew-brett:
Ariel - would you mind casting a quick eye over : https://github.com/nipy/nipy/pull/363? If you agree then we can close this one and get the fix merged soon.

PR #318, opened 2014-05-06
Flip left and right in plot_map.

csytracy:


matthew-brett:
Sorry to take so long to get back to you on this one.

Yes, an extra parameter to the plot_map function would be a good option - would you mind having a go?

matthew-brett:
Hi - @csytracy - if you are interested to get this done, we could merge it soon, we will probably release in a few days.

csytracy:
Sorry, I don't have time to do this now.

PR #321, opened 2014-07-30
Document 4d realignment speedup/multiscale parameters

alexis-roche:
Following a recent discussion on the nipy list initiated by Michael Hallquist, I added docstrings to describe the arguments involved in motion estimation.

### DIFF
--- a/nipy/algorithms/registration/groupwise_registration.py
+++ b/nipy/algorithms/registration/groupwise_registration.py
@@ -33,7 +33,7 @@ from .affine import Rigid, Affine
 from ._registration import (_cspline_transform,
                             _cspline_sample3d,
                             _cspline_sample4d)
-from .type_check import (check_type, check_type_and_shape)
+
 
 VERBOSE = os.environ.get('NIPY_DEBUG_PRINT', False)
 INTERLEAVED = None
@@ -246,7 +246,7 @@ class Realign4dAlgorithm(object):
         # The reference scan conventionally defines the head
         # coordinate system
         self.optimize_template = optimize_template
-        if not optimize_template and refscan == None:
+        if not optimize_template and refscan is None:
             self.refscan = 0
         else:
             self.refscan = refscan
@@ -840,7 +840,7 @@ class Realign4d(object):
         maxfun : int 
             Maximum number of function evaluations in maxfun.
         """
-        if between_loops == None:
+        if between_loops is None:
             between_loops = loops
         t = realign4d(self._runs,
                       affine_class=self.affine_class,
--- a/nipy/algorithms/registration/groupwise_registration.py
+++ b/nipy/algorithms/registration/groupwise_registration.py
@@ -36,7 +36,6 @@ from ._registration import (_cspline_transform,
 
 VERBOSE = os.environ.get('NIPY_DEBUG_PRINT', False)
 INTERLEAVED = None
-OPTIMIZER = 'ncg'
 XTOL = 1e-5
 FTOL = 1e-5
 GTOL = 1e-5
@@ -44,20 +43,6 @@ STEPSIZE = 1e-6
 SMALL = 1e-20
 MAXITER = 64
 MAXFUN = None
-BORDERS = 1, 1, 1
-REFSCAN = 0
-EXTRAPOLATE_SPACE = 'reflect'
-EXTRAPOLATE_TIME = 'reflect'
-
-LOOPS = 5  # loops within each run
-BETWEEN_LOOPS = 5  # loops used to realign different runs
-SPEEDUP = 5  # image sub-sampling factor for speeding up
-"""
-# How to tune those parameters for a multi-resolution implementation
-LOOPS = 5, 1
-BETWEEN_LOOPS = 5, 1
-SPEEDUP = 5, 2
-"""
 
 
 def interp_slice_times(Z, slice_times, tr):
@@ -204,8 +189,7 @@ class Realign4dAlgorithm(object):
                  transforms=None,
                  time_interp=True,
                  subsampling=(1, 1, 1),
-                 borders=BORDERS,
-                 optimizer=OPTIMIZER,
+                 optimizer='ncg',
                  optimize_template=True,
                  xtol=XTOL,
                  ftol=FTOL,
@@ -213,11 +197,11 @@ class Realign4dAlgorithm(object):
                  stepsize=STEPSIZE,
                  maxiter=MAXITER,
                  maxfun=MAXFUN,
-                 refscan=REFSCAN):
+                 refscan=0):
 
         self.dims = im4d.get_shape()
         self.nscans = self.dims[3]
-        self.xyz = make_grid(self.dims[0:3], subsampling, borders)
+        self.xyz = make_grid(self.dims[0:3], subsampling, (1, 1, 1))
         masksize = self.xyz.shape[0]
         self.data = np.zeros([masksize, self.nscans], dtype='double')
 
@@ -244,8 +228,8 @@ class Realign4dAlgorithm(object):
         # The reference scan conventionally defines the head
         # coordinate system
         self.optimize_template = optimize_template
-        if not optimize_template and refscan is None:
-            self.refscan = REFSCAN
+        if not optimize_template and refscan == None:
+            self.refscan = 0
         else:
             self.refscan = refscan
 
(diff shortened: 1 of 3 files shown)


nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -758,19 +741,71 @@ def _init(self, images, tr, slice_times, slice_info, affine_class):
...
+            Defines the optimization method. One of 'simplex',
+            'powell', 'cg', 'ncg', 'bfgs' and 'steepest'.
+        refscan : None or int
+            Defines the number of the scan used as the reference
+            coordinate system for each run. If None, a reference
+            coordinate system is defined internally that does not
+            correspond to any particular scan. Note that the
+            coordinate system associated with the first run is always
+            used for between-run motion estimation.
+        borders : sequence of ints
```
matthew-brett: Sorry - my question was bad - I think the order of parameters in the docstring does not match the signature (and didn't before this change either).  Is that right?

matthew-brett:
@alexis-roche - any progress here?

alexis-roche:
`borders` are tested as of https://github.com/nipy/nipy/commit/211e4cfd5dc019718fb6d00af1d8a9a5f791024e

alexis-roche:
@matthew-brett But you did well to spot this, because there was actually a bug when choosing `borders` too large compared to image shape

matthew-brett:
Do I understand right, that you are only testing if the code runs with a different 'border' parameter, but not whether it generates the right output?   Can you test whether the border parameter generates the right output?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -506,17 +494,17 @@ def adjust_subsampling(speedup, dims):
 def single_run_realign4d(im4d,
                          affine_class=Rigid,
                          time_interp=True,
-                         loops=LOOPS,
-                         speedup=SPEEDUP,
-                         borders=BORDERS,
-                         optimizer=OPTIMIZER,
+                         loops=5,
+                         speedup=5,
+                         refscan=0,
```
matthew-brett: Alexis - I think you've changed the order of parameters, in that `refscan` is now after `speedup` not after `maxfun`.   That's an API change (in case anyone is using say `borders` as a positional argument.  Any reason for that?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -563,8 +550,8 @@ def format_arg(x):
                                affine_class=affine_class,
                                time_interp=time_interp,
                                subsampling=subsampling,
-                               borders=borders,
                                refscan=refscan,
+                               borders=borders,
```
matthew-brett: Order changed here too?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -587,22 +574,21 @@ def realign4d(runs,
...
               align_runs=True,
-              loops=LOOPS,
-              between_loops=BETWEEN_LOOPS,
-              speedup=SPEEDUP,
-              borders=BORDERS,
-              optimizer=OPTIMIZER,
+              loops=5,
+              between_loops=5,
+              speedup=5,
+              refscan=0,
```
matthew-brett: Order changed here too.

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -629,6 +615,7 @@ def realign4d(runs,
                                        time_interp=time_interp,
                                        loops=loops,
                                        speedup=speedup,
+                                       refscan=refscan,
```
matthew-brett: Here too.

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -758,20 +744,89 @@ def _init(self, images, tr, slice_times, slice_info, affine_class):
...
     def estimate(self,
-                 loops=LOOPS,
+                 loops=5,
                  between_loops=None,
                  align_runs=True,
-                 speedup=SPEEDUP,
-                 borders=BORDERS,
-                 optimizer=OPTIMIZER,
+                 speedup=5,
+                 refscan=0,
```
matthew-brett: And here.

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -758,20 +744,89 @@ def _init(self, images, tr, slice_times, slice_info, affine_class):
...
                  maxiter=MAXITER,
-                 maxfun=MAXFUN,
-                 refscan=REFSCAN):
-        if between_loops is None:
+                 maxfun=MAXFUN):
+        """Estimate motion parameters.
+
+        Parameters
+        ----------
+        loops : int or sequence of ints
```
matthew-brett: I think all parameters should have `..., optional`

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -780,6 +835,7 @@ def estimate(self,
                       loops=loops,
                       between_loops=between_loops,
                       speedup=speedup,
+                      refscan=refscan,
```
matthew-brett: Order change again.

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -198,6 +198,8 @@ def test_spacetimerealign_params():
     assert_array_equal(R1.slice_times, np.arange(3) / 3. * 2.)
     # Smoke test run
     R1.estimate(refscan=None, loops=1, between_loops=1, optimizer='steepest')
+    # Test tighter borders for motion estimation
+    R1.estimate(refscan=None, loops=1, between_loops=1, optimizer='steepest', borders=(5,5,5))
```
matthew-brett: Sorry to be boring, but can you think of a test of the output here?

matthew-brett:
Alexis - will you have time to get to this in the next couple of days?

matthew-brett:
Alexis - I guess you are on holiday or something.  If I don't hear from you in the next 24 hours, I'll take the liberty of switching the order of the parameters back to their original order, and then merging.

alexis-roche:
Hey Matthew, sorry for being very slow to get back to this...

I changed the parameter order because it did not match the docstring, and I thought that the docstring order made much more sense. I didn't think it to be a critical API break because those are kw args and did not expect any one to use kw args as positional ones. But, on a more general stance, I have been working under the assumption that API breakages are  acceptable until we reach version >= 1. That is something we need to clarify. 

In the meantime, I wouldn't support going back to the initial order as it would not make much sense.  Would a warning be ok?

matthew-brett:
I think it's OK to break API, but only if the result is that someone using the old API gets a new error message telling them what to do.  I don't think we should make API changes giving a silent change in the meaning of code written for the old API.

In this case, I'm sure you're right that it's unlikely anyone has used `borders` in this case as a positional argument.  But, just to be safe, maybe you could put a check that `refscan` and `borders` are the shape you expect, and raise an error explaining about the change in parameter order if not?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -213,10 +198,13 @@ def __init__(self,
                  stepsize=STEPSIZE,
                  maxiter=MAXITER,
                  maxfun=MAXFUN,
-                 refscan=REFSCAN):
+                 refscan=0):
```
matthew-brett: But you have the old order here?

alexis-roche:
@matthew-brett Hope this addresses your concerns

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -27,16 +27,16 @@
...
                                         as_xyz_image)
 from ..slicetiming import timefuncs
+from .type_check import (check_type, check_type_and_shape)
 from .optimizer import configure_optimizer, use_derivatives
 from .affine import Rigid, Affine
 from ._registration import (_cspline_transform,
                             _cspline_sample3d,
                             _cspline_sample4d)
-
+from .type_check import (check_type, check_type_and_shape)
```
matthew-brett: Duplicate import line

nipy/algorithms/registration/type_check.py
```diff
@@ -0,0 +1,46 @@
+"""
```
matthew-brett: Maybe this should be a private module?

Consider adding docstrings to public funcs?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -244,8 +246,8 @@ def __init__(self,
         # The reference scan conventionally defines the head
         # coordinate system
         self.optimize_template = optimize_template
-        if not optimize_template and refscan is None:
-            self.refscan = REFSCAN
+        if not optimize_template and refscan == None:
```
matthew-brett: This should be `refscan is None` : https://www.python.org/dev/peps/pep-0008/#programming-recommendations

matthew-brett:
Thanks - I think we got there in the end.  Sorry to be slow to get to this one.

PR #322, opened 2014-09-13
RF: move scriptrunner code into own module

matthew-brett:
Use scriptrunner module that started off in the delocate project. Well, it
started in the test_scripts.py module here, but I refactored it in the
delocate project, and the refactoring is coming back home.


matthew-brett:
Any comments here?

A version of this change went into dipy, nibabel, delocate already, so I will merge soon unless I hear otherwise.

PR #334, opened 2015-06-06
MRG: move doc building to use gh-pages

matthew-brett:
Machinery to upload docs to gh-pages.

Deprecate old sourceforge stuff.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -14,7 +14,7 @@ PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -d build/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 
-.PHONY: help clean pdf all dist public_html web web_public htmlonly api html pickle htmlhelp latex changes linkcheck doctest
+.PHONY: help clean pdf all dist htmlonly api html pickle htmlhelp latex changes linkcheck doctest
 
 help:
 	@echo "Please use \`make <target>' where <target> is one of"
@@ -30,9 +30,6 @@ help:
 	@echo "  doctest   	run doctests in reST files"
 	@echo "  pdf		make <latex> and run the PDF generation"
 	@echo "  dist		make <all> and put results in $DIST_DIR/"
-	@echo "  public_html	make <dist> and rsync with public website"
-	@echo "  upload-stable-web	make <dist> and rsync with public stable website"
-	@echo "  upload-devel-web	make <dist> and rsync with public devel website"
 	@echo "  gitwash-update	update git workflow from source repo"
 
 # Commented these out, wasn't clear if we'd use these targets or not.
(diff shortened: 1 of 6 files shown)


doc/Makefile
```diff
@@ -56,19 +53,23 @@ dist: clean all
 	@echo "Build finished.  Final docs are in $(DIST_DIR)"
 
 # This one udates for the specific user named at the top of the makefile
+# Please don't use this - we use github pages now
```
bcipolli: Why not delete these lines?

bcipolli:
@matthew-brett can you give a list of commands to test this locally? I think it is:

```
make html
./tools/upload-gh-pages.sh doc/dist nipy bcipolli
```

But I want to make sure before running things here. When I start seeing `rm -rf .git`, I like to double-check :)

bcipolli:
@matthew-brett last question--is the travis build broken on the master branch? I see two failures here, but I can't imagine any change here causes them. 

Perhaps this will be fixed if you rebase--maybe they're fixed in the latest code?

matthew-brett:
Ben - thanks for the review - as ever.

The command that should work is `make upload-html` 

I think the test errors aren't related, I'm working on these at the moment.

doc/Makefile
matthew-brett: OK - done.

bcipolli:
@matthew-brett I want to test in a way that will allow me to view the results. I believe `make upload-html` will try to push this to the nipy org's github page, rather than my own.

matthew-brett:
Ah - sorry - I see what you mean - you were right first time:

```
make html
./tools/upload-gh-pages.sh doc/dist nipy bcipolli
```

tools/upload-gh-pages.sh
```diff
@@ -0,0 +1,29 @@
...
+if [ -z "$PROJECT" ]; then
+    echo $USAGE
+    exit 1
+fi
+ORGANIZATION=$3
+if [ -z "$ORGANIZATION" ]; then
+    ORGANIZATION=nipy
+fi
+upstream_repo="https://github.com/$ORGANIZATION/$PROJECT"
+cd $HTML_DIR
```
bcipolli: Can you make sure this command succeeds, or that the `$HTML_DIR` exists? This is the major source of user error for me... and though this command fails, something gets uploaded (but the wrong thing).

bcipolli:
@matthew-brett I was able to get this to work by running:
```
cd doc
make html
../tools/upload-gh-pages.sh build/html nipy bcipolli
```
A few comments:
* If I point the `HTML_DIR` to something that doesn't exist, the wrong thing is uploaded. A check to make sure the directory exists would be appreciated (I can imagine pushing to my own repo by this method first, to make sure things work before pushing to `nipy`, for example)
* I find the second two parameters to be a bit unintuitive. In git, we usually specify the remote name first, then the repo name second.

Otherwise, seems to work great! No need to change the above, but for manual use, it would make things easier.

tools/upload-gh-pages.sh
matthew-brett: Good point - sorry - I had already done this in the other projects, but forgot to upload these changes here.  How about now?

bcipolli: Much better. Turns out that my root `.git` directory was deleted because the `cd $HTML_DIR` failed and I was running in the root directory. Re-cloning now ... but that was less than ideal :)

bcipolli: OK, re-tested and this works great.

matthew-brett:
@bcipolli - OK to merge this one?

bcipolli:
yes please! :)

PR #332, opened 2015-03-26
MRG: Tried to fix failing tests due to numerical issues

bthirion:
Trying to solve the Travis errors. I don't reproduce them on my box, which makes the whole thing a bit complicated. I am assuming that these are simply numerical issues.

Disclaimer: For some reason, I don't get Travis output atm.

PR #339, opened 2015-08-24
Fixing tests

bcipolli:
There were two test failures that break the build. I believe the fixes I made are correct:

* Use numpy's `assert_array_almost_equal` instead of `assert_almost_equal` on a 3D array.
* matrix rank should be the same (10) whether tolerance is +/- eps

It'd be great to get this merged in, looks like Python 2.x build has been broken for some time.

### DIFF
--- a/nipy/algorithms/utils/tests/test_matrices.py
+++ b/nipy/algorithms/utils/tests/test_matrices.py
@@ -31,7 +31,7 @@ def test_matrix_rank():
     eps = np.finfo(X.dtype).eps
     assert_equal(matrix_rank(X, tol=0), 10)
     assert_equal(matrix_rank(X, tol=S.min() - eps), 10)
-    assert_equal(matrix_rank(X, tol=S.min() + eps), 9)
+    assert_equal(matrix_rank(X, tol=S.min() + eps), 10)
 
 
 def test_full_rank():
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,7 @@
 language: python
 python:
     - 2.6
-    - 3.2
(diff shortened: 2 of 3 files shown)


bthirion:
I think I had addressed those in #332, but all right, travis is green  again. I'm going to merge.
Note however that this code base is bound to die soon anyway...

nipy/algorithms/utils/tests/test_matrices.py
```diff
@@ -31,7 +31,7 @@ def test_matrix_rank():
     eps = np.finfo(X.dtype).eps
     assert_equal(matrix_rank(X, tol=0), 10)
     assert_equal(matrix_rank(X, tol=S.min() - eps), 10)
-    assert_equal(matrix_rank(X, tol=S.min() + eps), 9)
+    assert_equal(matrix_rank(X, tol=S.min() + eps), 10)
```
matthew-brett: This test was correct - have a look at the `matrix_rank` code.  The test checks that if the tolerance is less than the `S.min()`, then the rank drops by one.

matthew-brett: Sorry - if the tolerance is greater than `S.min()`.

matthew-brett: On my laptop and a couple of linux boxes I have access too, the edited version of the test fails, as I would expect - so I am not sure what's happening with the travis-ci test.

matthew-brett: Test fixed here : https://github.com/nipy/nipy/pull/343

PR #341, opened 2015-08-25
MRG: refactor, fix traxis build

matthew-brett:
Python 3 builds broken for travis - refactor using dipy as template to get
better handle on the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,21 +4,34 @@
 # It's okay for it to be on multiple physical lines, so long as you remember:
 # - There can't be any leading "-"s - All newlines will be removed, so use
 # ";"s
+sudo: false  # To use travis container infrastructure
 language: python
+cache:
+  directories:
+    - $HOME/.cache/pip
+env:
+    global:
+        - DEPENDS="numpy scipy sympy matplotlib nibabel"
 python:
     - 2.6
+    - 3.2
+    - 3.3
+    - 3.4
 
 matrix:
   include:
     - python: 2.7
       env:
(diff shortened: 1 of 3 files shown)


matthew-brett:
I think this one is ready to merge.  The Python 3 builds give errors, but the errors now show the real problem rather than an unrelated build problem, as we had before.

The new minimal dependency check also fails, but in a useful way, because Alexis' changes have pushed our scipy dependency up about 0.12, which needs some thought.

matthew-brett:
Any comments here?

bthirion:
Sorry, I don,'t understand what you want to do here. The build is broken again, even under Python 2.

matthew-brett:
I tried to explain in the comments.

The standard Python 2 builds pass, as before.

The Python 3 builds fail, but usefully, instead of failing at the build step.

I added an extra Python 2 check against our declared minimum dependencies, which fails because of the Lapack refactor, which we are currently reviewing.

matthew-brett:
Any comments here?   If not, I'll merge soon.

bcipolli:
I think we need to merge this one ASAP, to better see impact on #346, #348 and subsequent PRs. The biggest change of note for me is to a minimum sympy from 0.6.6 to 0.7. 

As long as that's cool, I suggest merge so we can get down to investigating test failures in all PRs.

matthew-brett:
I think the sympy change is correct - I had just forgotten to update that earlier.

I agree - let's merge this one to make it easier to test the rest.  I'll rebase the rest after merging in this one.

PR #343, opened 2015-08-25
BF: restore and fix original matrix rank test

matthew-brett:
Test was testing if increasing the threshold resulted in lower matrix rank
scores.  This was the right test, but we were using numpy.linalg.svd to get
the eigenvalues, rather than scipy.linalg.svd, so the values were slightly
different from the ones seen inside the tested routine, and the test was
failing.


bthirion:
LGTM. Merging

Issue #344, opened 2015-08-26
'remove lapack linking' merge forces scipy dependency up to 0.13


matthew-brett:
See `nipy.labs` errors in https://travis-ci.org/matthew-brett/nipy/jobs/77317008

We previously depended on scipy >= 0.7.0, but I think that the remove lapack merge means that we now need scipy 0.13.0 which is much more recent.  This is in order to use `scipy.linalg._fblas` (arrived in scipy 0.12) and then `scipy.linalg._fblas.dsymm` (arrived in scipy 0.13).

The question is, what to do now.  Options I can see:

* revert the 'remove lapack linking' merge on the basis that the convenience of not linking to external lapack is less then the convenience of being able to use scipy < 0.13;
* put some import checks around the nipy.labs and nipy.labs testing code that skips the relevant tests when the scipy version is too low.

Alexis - what do you think?

alexis-roche:
I think that we should minimize efforts here given our limited manpower and the fact that the nipy package itself is getting deprecated... So I think it's fine to revert to the old, historical "lapack needed" situation.

matthew-brett:
OK - what I propose to do for now, is to revert the 'remove lapack' merge, and reopen the contents as a PR. 

Any comments before I go ahead with that?

satra:
@alexis-roche, @matthew-brett - since this merge was intended to simplify install of nipy in the context of  mkl, it affects users who use conda + mkl as their platform. i'm assuming that reverting this merge will affect such an environment. 

if so, is the plan to tell users to shift over to nireg and update their scripts?

matthew-brett:
Satra - I think our priority now is to get down to no test errors.   So I propose to revert this merge and re-open the original PR with the intention of merging again soon, when we can do that without test errors.  Sound reasonable to you - Satra, Alexis?

satra:
@matthew-brett - from a test perspective that sounds fine. and the pr solution would be to allow for skipping tests on (scipy < 0.13)?

matthew-brett:
Skipping the tests for scipy < 0.13 is one option, if we can't find a better one.

Is the mkl a big issue for you?   I guess we'll do a nipy release in the next few weeks.

Alexis - is there another way of working round the mkl issue that you can think of?

satra:
[...]

Issue #342, opened 2015-08-25
Linking errors, probably from "remove lapack linking" branch


matthew-brett:
Before we merged #328 - we weren't getting any link errors - see https://travis-ci.org/nipy/nipy/builds/56913971

After we merged #328 - the Python 3 build broke down, probably for unrelated reasons - see #338 . 

Re-enabling the Python 3 build (#341) we now have 12 linking errors - see : https://travis-ci.org/matthew-brett/nipy/jobs/77184112

Errors are of form:

```
ImportError: /home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/labs/utils/routines.cpython-34m.so: undefined symbol: PyCObject_AsVoidPtr
```

This is only for Python 3 - no link errors for Python 2:

https://travis-ci.org/matthew-brett/nipy/jobs/77184109

@alexis-roche - I think these are from the lapack merge - is that possible?

alexis-roche:
Oh, yes, this is due to a change in the Python C API from 2.7 to 3. My code uses Python CObjects, which turn out to be deprecated as of Python 2.7, and are probably no longer supported in Python 3 (https://docs.python.org/2/c-api/cobject.html). I think I see how to fix this using the new "PyCapsule" object, and will submit a patch later this week.

matthew-brett:
Thanks - that would be very helpful...

PR #328, opened 2014-11-11
Remove lapack linking

alexis-roche:
This PR implements an alternate strategy to bind BLAS/LAPACK routines based on importing functions at run time via C-pointers provided by scipy. I removed any attempt to link BLAS/Lapack libraries at build time, even using source code (the lapack-lite distribution therefore becomes obsolete).

Because not all the BLAS/LAPACK functions are available through scipy, I had to slightly rewrite some parts of the fff library and the `labs` package. The tests are however unchanged and all pass on my box.

Issue #258, opened 2013-02-03
Seemingly random failures in formula


matthew-brett:
Testing on Python 3.3:

<pre>
======================================================================
ERROR: nipy.algorithms.statistics.formula.tests.test_formula.test_formula_from_recarray
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/numpy/core/fromnumeric.py", line 1507, in sum
    sum = a.sum
AttributeError: 'list' object has no attribute 'sum'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/nose-1.2.1-py3.3.egg/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/buildslave/osx-10.6/nipy-bdist-mpkg-3_3/build/venv/lib/python3.3/site-packages/nipy-0.3.0.dev-py3.3-macosx-10.6-intel.egg/nipy/algorithms/statistics/formula/tests/test_formula.py", line 133, in test_formula_from_recarray
    f = F.Formula.fromrec(D, drop='y')
  File "/Users/buildslave/osx-10.6/nipy-bdist-mpkg-3_3/build/venv/lib/python3.3/site-packages/nipy-0.3.0.dev-py3.3-macosx-10.6-intel.egg/nipy/algorithms/statistics/formula/formulae.py", line 538, in fromrec
    return np.sum(list(f.values()))
  File "/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/numpy/core/fromnumeric.py", line 1510, in sum
    out=out, keepdims=keepdims)
  File "/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/numpy/core/_methods.py", line 18, in _sum
    out=out, keepdims=keepdims)
  File "/Users/buildslave/osx-10.6/nipy-bdist-mpkg-3_3/build/venv/lib/python3.3/site-packages/nipy-0.3.0.dev-py3.3-macosx-10.6-intel.egg/nipy/algorithms/statistics/formula/formulae.py", line 587, in __add__
    f = self.__class__(np.hstack([self.terms, other.terms]))
TypeError: __init__() missing 1 required positional argument: 'levels'

======================================================================
[...]

PR #346, opened 2015-08-27
MRG: temporarily revert lapack refactor for testing

matthew-brett:
Lapack refactor increases our scipy dependency to 0.13 (see gh-344) and causes
some build errors (see gh-342).

Revert for now, so we can reconsider how best to fix up gh-328.

As discussed in gh-344.


matthew-brett:
OK - I think this one is ready now - Alexis - do you have time to take a look?

matthew-brett:
@alexis-roche - would you mind having a look at this one?   It's the last big piece in fixing the builds.

matthew-brett:
If anyone is reviewing this - it is best reviewed by commit, not via the "Files changed" interface.

The first commit 9d83fc3 just reverts Alexis' previous changes to make the code independent of an external BLAS  / LAPACK library, so doesn't need a review (other than maybe thinking about the policy of reverting it, discussed in the intro at the top of this PR).

The second commit d9a2983 is a tiny Makefile change to help compile the Cython .pyx files.

The third commit d5b81f6 is just a rebuild of a .c file from the .pyx file and doesn't need review.

The fourth commit cd595cf does need some review, because it's a change in the way of sharing the Numpy C API table.  This commit doesn't contain the changes from the .c files generated from the .pyx changes, to help with review.

The last commit 14e5332 is just the generated .c file changes, and doesn't need review.

Thanks - sorry this one is a little annoying to review.

effigies:
I had a look. Second and fourth commits look good.

Remaining errors seem to be #258 related?

matthew-brett:
Yes, the last errors should be fixed in nipy master (which has the fix for #258 .   I tested an integration branch here (current master with this PR): https://travis-ci.org/matthew-brett/nipy/builds/77797820

effigies:
Cool. Looks good, then, noting that I didn't try to look at the reversion commit.

matthew-brett:
Thanks for the review.

Will hold off merge until Monday hoping that Alexis has time to take a look.

alexis-roche:
Yes, sorry I am off for the week-end with just a smartphone. I'll have a
look first thing Monday morning. Cheers.

alexis-roche:
OK, I see that it's reverted to the point where "lapack lite" is compiled by default, so at least no linking with a system lapack is attempted (this can only be changed by manually editing `nipy/labs/setup.py`). It works on my box. 

I think we here have a robust solution to bind the required lapack/blas routines, one that works with many scipy and python versions. It would in theory be more elegant to use the C function pointers exposed by the scipy API, however in practice it would be a mess if we jointly consider scipy < 0.13 and python > 2.7. My gut feeling is to stick to the pragmatic "lapack lite" approach, given that nipy is becoming obsolete anyway.

matthew-brett:
So I think you're saying you think this one is good to merge?

Should we provide a configuration option to link to optimized lapack if available (instead of requiring an edit to `nipy.labs.setup.py`)?

alexis-roche:
Yes, sorry, good to merge. I thought about providing a config option, but didn't see a standard way to do this at the time. What we want to do is reset the `WANT_LAPACK_LINKING` variable in `nipy.labs.setup.py`, which defaults to False.

matthew-brett:
OK - let's merge for now, and I'll have a look to see if I can find some configuration option in shortish exploration time.

matthew-brett:
Here's my attempt at the configuration stuff : https://github.com/nipy/nipy/pull/352

PR #348, opened 2015-08-28
MRG: fix intermittent errors from Formula

matthew-brett:
Closes gh-258.

Intermittent errors testing Formula, turned out to be from:

* An error I introduced trying to make it easier to use sub-classes with
  Formula add and subtract;
* Unpredictable order of terms after subtraction, caused by use of set to find
  unique terms.

### DIFF
--- a/nipy/algorithms/statistics/formula/formulae.py
+++ b/nipy/algorithms/statistics/formula/formulae.py
@@ -180,13 +180,13 @@ class Term(sympy.Symbol):
 
     def _getformula(self):
         return Formula([self])
-    formula = property(_getformula, doc="Return a Formula with only terms=[self].")
+    formula = property(_getformula,
+                       doc="Return a Formula with only terms=[self].")
 
     def __add__(self, other):
         if self == other:
             return self
-        else:
-            return sympy.Symbol.__add__(self, other)
+        return sympy.Symbol.__add__(self, other)
 
 
 # time symbol
@@ -416,10 +416,10 @@ class Formula(object):
     model being the sum of each term multiplied by a linear regression
     coefficient.
 
-    The expressions may depend on additional Symbol instances,
-    giving a non-linear regression model.
+    The expressions may depend on additional Symbol instances, giving a
+    non-linear regression model.
     """
-    # This flag is defined to avoid using isinstance
+    # This flag is defined for test isformula(obj) instead of isinstance
     _formula_flag = True
 
     def __init__(self, seq, char = 'b'):
@@ -558,10 +558,20 @@ class Formula(object):
         return self.__class__([term.subs(old, new) for term in self.terms])
 
     def __add__(self, other):
-        """
-        Create a new Formula by combining terms
-        of other with those of self.
+        """ New Formula combining terms of `self` with those of `other`.
 
+        Parameters
+        ----------
+        other : Formula instance
+            Object for which ``is_formula(other)`` is True
+
+        Returns
+        -------
+        added : Formula instance
+            Formula combining terms of `self` with terms of `other`
+
+        Examples
+        --------
         >>> x, y, z = [Term(l) for l in 'xyz']
         >>> f1 = Formula([x,y,z])
         >>> f2 = Formula([y])+I
@@ -579,11 +589,23 @@ class Formula(object):
         return f
 
     def __sub__(self, other):
-        """
-        Create a new Formula by deleting terms in other
-        from self. No exceptions are raised for terms in other that do not appear in
-        self.
+        """ Formula from deleting terms in `other` from those in `self`
 
+        No exceptions are raised for terms in `other` that do not appear in
+        `self`.
+
+        Parameters
+        ----------
+        other : Formula instance
+            Object for which ``is_formula(other)`` is True
+
+        Returns
+        -------
+        subbed : Formula instance
+            Formula with terms of `other` removed from terms of `self`
+
+        Examples
+        --------
         >>> x, y, z = [Term(l) for l in 'xyz']
         >>> f1 = Formula([x,y,z])
         >>> f2 = Formula([y])+I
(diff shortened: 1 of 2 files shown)


nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -575,33 +585,47 @@ def __add__(self, other):
...
-        f = self.__class__(np.hstack([self.terms, other.terms]))
+        f = Formula(np.hstack([self.terms, other.terms]))
         return f
 
     def __sub__(self, other):
-        """
-        Create a new Formula by deleting terms in other
-        from self. No exceptions are raised for terms in other that do not appear in
-        self.
+        """ Formula from deleting terms in `other` from those in `self`
```
yarikoptic: have difficulty parsing this sentence ... I did understand previous form ;)

nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -575,33 +585,47 @@ def __add__(self, other):
         """
         if not is_formula(other):
             raise ValueError('only Formula objects can be added to a Formula')
-        f = self.__class__(np.hstack([self.terms, other.terms]))
+        f = Formula(np.hstack([self.terms, other.terms]))
```
yarikoptic: so now even subclasses would return Formula.  I am not familiar much with this codebase, but is that desired? why in previous form it was faulty?

nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -575,33 +585,47 @@ def __add__(self, other):
...
         _b0*x + _b1*z
         """
         if not is_formula(other):
-            raise ValueError('only Formula objects can be subtracted from a Formula')
-        d = list(set(self.terms).difference(other.terms))
-        return self.__class__(d)
+            raise ValueError(
+                'only Formula objects can be subtracted from a Formula')
+        # Preserve order of terms in subtraction
+        d = [term for term in self.terms if term not in other.terms]
```
bcipolli: Here is the real change, from set (order lost) to list filtering (order maintained)

nipy/algorithms/statistics/formula/tests/test_formula.py
```diff
@@ -43,24 +43,24 @@ def test_terms():
 def test_getparams_terms():
     t = F.Term('t')
     x, y, z = [sympy.Symbol(l) for l in 'xyz']
-    yield assert_equal, set(F.getparams(x*y*t)), set([x,y])
-    yield assert_equal, set(F.getterms(x*y*t)), set([t])
+    assert_equal(set(F.getparams(x*y*t)), set([x,y]))
+    assert_equal(set(F.getterms(x*y*t)), set([t]))
```
yarikoptic: I love and hate generators in the tests... why previous form was "incorrect"?  did nose (or whatever) didn't pick up this test as a "test generator" or what?

nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -575,33 +585,47 @@ def __add__(self, other):
...
+        return Formula(d)
```
bcipolli: Why not keep the `self.__class__` logic? I don't think I know Python enough to really understand the impact, but isn't that safer / more correct if `Formula` is subclassed?

nipy/algorithms/statistics/formula/formulae.py
matthew-brett: OK - will rewrite.

nipy/algorithms/statistics/formula/formulae.py
matthew-brett: The problem was that the add and sub methods create a new object at the end, from a list of terms.  By using ``self.__class__`` to construct the output object, I am asserting that the class constructor has the same signature as the Formula class, and that is not true of Factor, which needs a name.

I guess one way round that would be to have a ``from_terms`` method to Formula, and use that, but in fact I don't think its unreasonable to return a Formula from the Factor subtraction, the main logic of the Factor is in the constructor, and combining two factors can make something that isn't a pure Factor

nipy/algorithms/statistics/formula/tests/test_formula.py
matthew-brett: I've been gradually removing test generators from the code because the errors are harder to debug.  Here I did it as part of refactoring to get ready to add the new tests, no special reason.

nipy/algorithms/statistics/formula/formulae.py
matthew-brett: See https://github.com/nipy/nipy/pull/348#discussion_r38202480 - does that answer your question?

bcipolli: Sounds good!

matthew-brett:
Guys - have I answered your comments?

yarikoptic:
as to me -- yes ;)  I will keep one more whining shortly

nipy/algorithms/statistics/formula/formulae.py
yarikoptic: there was a time when python folks promised the death of filter superseeded by list comprehensions.... everyone shifted but I find myself using filter/map more since they are easier to read
```python
d = filter(lambda term: term not in other.terms, self.terms)
```
the only thing is that this one is not python3-compatible so if you use six you would need to
```
from six.moves import filter
d = list(filter(lambda term: term not in other.terms, self.terms))
```

yarikoptic: but then you could also use
```python
from six.moves import filterfalse
d = list(filterfalse(other.terms.__contains__, self.terms))
```
I guess ;)

yarikoptic:
Travis is not happy though yet

matthew-brett:
Yarik - forgive my rustic taste - but I prefer the list comprehension.   Can you accept your brother coder, even with his faults?

matthew-brett:
I'm afraid travis will continue to fail here because other problems (with the LAPACK stuff).  The tests should pass when https://github.com/nipy/nipy/pull/346 is in and this is rebased on top.  But the Python 3 tests of these changes do pass.

yarikoptic:
I have deep respect to seniors' tastes :-) whatever you feel more comfortable with!

matthew-brett:
OK - I think that agreement from y'all, merging.

PR #349, opened 2015-08-28
RF: replace == None with is None checks

matthew-brett:
We were getting lots of warnings about use of == None or != None. Replace all
instances of == None with "is None", != None with "is not None"


satra:
LGTM

bcipolli:
Agreed. please merge if the Travis build completes successfully!

effigies:
The preferred style for a lot of these would be `x is not None` rather than `not x is None`.

matthew-brett:
Chris - that's true - here I was only doing the simplest possible search and replace to see if it would work.  Should I go back through and apply that style do you think?

effigies:
Reading through, I was assuming this was a style issue, but looking again it's warnings. If it doesn't produce warnings and would be a serious time sink, I wouldn't worry, since they produce the same byte code.

matthew-brett:
Actually it turned out to be trivial to script with `perl -p -i -e` : - see 7ee6d24

bcipolli:
Nice. I find extraneous warnings to be very dangerous--makes it hard to detect meaningful warnings. So glad you made the change, @matthew-brett --thanks!

effigies:
All looks reasonable to me.

bthirion:
LGTM

matthew-brett:
OK - merging.

Issue #325, opened 2014-10-22
plot_map error with numpy 1.9


rdevon:
It appears that in:
labs/datasets/volumes/volume_img.py line 280:
pixdim = np.diag(A)
needs to be
pixdim = np.diagonal(A).copy()

"pixdim[0] = -pixdim[0]" threw a warning on earlier versions of numpy (1.8), but breaks outright in 1.9.

crcox:
Just wanted to mention that this problem seems to still exist. I installed NiPy with pip and ran into exactly rdevon's error. Installed from the master branch on the repo, and the error goes away.

GaelVaroquaux:
Hi,

We are progressively advising people to use the nilearn plotting
function, that are similar to the nipy ones, but benefit from a more
modern codebase and more maintenance.
http://nilearn.github.io/auto_examples/index.html#manipulating-images-and-visualization
http://nilearn.github.io/building_blocks/plotting.html

If you find any problems with these, don't hesitate to report problems on
the nilearn issue tracker.

rdevon:
crcox: if you need it to work under nipy, with pip install: 
pip install git+https://github.com/nipy/nipy.git

crcox:
Sorry for never coming back to thank you both for the advice! Best.

bcipolli:
This error still exists. Either it needs to be fixed, or this function needs to be deprecated and/or removed. This is a blocking bug--severe!

If I submit a PR, will someone review it?

bthirion:
I can review it if needed.

bcipolli:
Sorry for the confusion. This looks like it was fixed by @GaelVaroquaux  in 548f0890813a11ba5edf9239bae2ececc3549bd3 . Just hasn't been released.

matthew-brett:
Fixed in master I believe - so closing - please re-open if you disagree.

PR #350, opened 2015-08-28
diag = >diag.copy() when the matrix will be manipulated.

bcipolli:
This is a general fix for #325 . In the newest version of `numpy`, when `diag` returns a vector of numbers along the diagonal, that vector is read-only. In places where we wish to manipulate the elements of that vector, we need to call `copy()` first.

I manually went through to check each call to `diag`, and to see where we manipulated the elements or returned the vector value.

**Questions:**

* Do we have tests that use `numpy==1.9`? 
* Do we want to add a specific plotting regression for #325?

### DIFF
--- a/nipy/labs/viz_tools/maps_3d.py
+++ b/nipy/labs/viz_tools/maps_3d.py
@@ -56,13 +56,13 @@ def affine_img_src(data, affine, scale=1, name='AffineImage',
         # Try out old install of Mayavi, with namespace packages
         from enthought.mayavi.sources.api import ArraySource
     center = np.r_[0, 0, 0, 1]
-    spacing = np.diag(affine)[:3]
+    spacing = np.diag(affine)[:3].copy()
     origin = np.dot(affine, center)[:3]
     if reverse_x:
         # Radiologic convention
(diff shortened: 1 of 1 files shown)


matthew-brett:
The travis tests use latest numpy, as well as minimal numpy - is that enough?

If you feel like adding a plotting regression test, that would be very helpful.

bcipolli:
OK, then we should beef up the test coverage before merging this PR--there should be multiple issues, not just the plotting one.

I'll take a look at that now.

bthirion:
This PR looks good.

matthew-brett:
Ben - where are we with this one?

bcipolli:
Oops, I got sidetracked. I determined two of the changes were unnecessary, and that two of the changes are untested. I simply need to write tests for them.

I will do that now.

matthew-brett:
Seems to have another PR In list of commits?   Bad rebase maybe?

bcipolli:
Re- rebased. Travis build is looking fine.

matthew-brett:
Nice, thanks.

PR #351, opened 2015-08-28
Merge maint/0.3.x into master

bcipolli:
These look like good changes that weren't merged into the current master. I had based my fix for #325 on this branch, for back compatibility purposes, and was surprised to find commits in here but not in master.


matthew-brett:
Thanks

PR #352, opened 2015-08-31
CONFIG: add config option for LAPACK compile

matthew-brett:
Add environment variable and setup.cfg option for compiling against external
lapack.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@
 # - There can't be any leading "-"s - All newlines will be removed, so use
 # ";"s
 sudo: false  # To use travis container infrastructure
+addons:
+  apt:
+    packages:
+        - libblas-dev
+        - liblapack-dev
 language: python
 cache:
   directories:
@@ -27,6 +32,10 @@ matrix:
     - python: 2.7
       env:
         - DEPENDS="numpy==1.6.0 scipy==0.7.0 sympy==0.7.0 nibabel==1.2.0"
+    # Test compiling against external lapack
+    - python: 3.4
+      env:
+        - NIPY_EXTERNAL_LAPACK=1
 before_install:
     - source tools/travis_tools.sh
     - virtualenv --python=python venv
@@ -34,7 +43,6 @@ before_install:
     - python --version # just to check
     - retry pip install nose # always
     - wheelhouse_pip_install $DEPENDS
-    # - sudo apt-get install -qq libatlas-dev libatlas-base-dev gfortran libpng-dev
     - if [ "${COVERAGE}" == "1" ]; then
       pip install coverage;
       pip install coveralls;
--- a/nipy/algorithms/clustering/setup.py
+++ b/nipy/algorithms/clustering/setup.py
@@ -7,21 +7,6 @@ def configuration(parent_package='',top_path=None):
     config = Configuration('clustering', parent_package, top_path)
     config.add_subpackage('tests')
 
-    # We need this because libcstat.a is linked to lapack, which can
-    # be a fortran library, and the linker needs this information.
-    from numpy.distutils.system_info import get_info
-    lapack_info = get_info('lapack_opt',0)
-    if 'libraries' not in lapack_info:
-        # But on OSX that may not give us what we need, so try with 'lapack'
-        # instead.  NOTE: scipy.linalg uses lapack_opt, not 'lapack'...
-        lapack_info = get_info('lapack',0)
-
-    #config.add_extension(
-    #            '_clustering',
-    #            sources=['clustering.c'],
-    #            libraries=['cstat'],
-    #            extra_info=lapack_info,
-    #            )
     return config
 
 
--- a/nipy/labs/bindings/setup.py
+++ b/nipy/labs/bindings/setup.py
@@ -1,29 +1,17 @@
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
-import os, sys
 
 def configuration(parent_package='',top_path=None):
     from numpy.distutils.misc_util import Configuration
     config = Configuration('bindings', parent_package, top_path)
     config.add_subpackage('tests')
     config.add_subpackage('benchmarks')
-    # We need this because libcstat.a is linked to lapack, which can
-    # be a fortran library, and the linker needs this information.
-    from numpy.distutils.system_info import get_info
-    lapack_info = get_info('lapack_opt',0)
-    if 'libraries' not in lapack_info:
-        # But on OSX that may not give us what we need, so try with 'lapack'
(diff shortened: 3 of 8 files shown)


satra:
@matthew-brett - i can check this against mkl later tonight, but if you download miniconda and then `conda install accelerate`, you should be able to as well. it will give you a 30 day trial.

matthew-brett:
Alexis - I changed the priority of the setup file and env var - OK now?

matthew-brett:
Satra - thanks - please do get back to me if you do get time to test.

satra:
@matthew-brett - current master and this branch still break with mkl. the problem is that numpy's get lapack_info picks up the mkl link library that is only available for developing with mkl. i've posted an issue with continuum.

https://github.com/ContinuumIO/anaconda-issues/issues/431

i'll probably still update my pull-request/branch for those who want to install nipy in mkl environments.

matthew-brett:
Satra - do you mean that you get a build error when you set external lapack linking on, with the environment variable or the setup.cfg file?  I assume nipy builds correctly in the default state (no env var or edits to setup.cfg)?

alexis-roche:
@matthew-brett 
I see that at line 74 of `nipy/labs/setup.py` and onwards, you are calling numpy's lapack info routine, while it's not needed if the user does not want system lapack linking. This might cause the crash reported by @satra . I think you want to embed that call in the loop further down. Sorry for not spotting that yesterday.

matthew-brett:
Surely calling `get_info` should not cause a crash if lapack is not found?

Satra - what error are you getting?

satra:
@matthew-brett 

in the mkl setting there is an external lapack available as far as numpy is concerned, i.e. mkl. and then it picks up libraries as shown in the continuum issue that cannot be linked against (see error below).

i believe @alexis-roche is correct in that if we are to rely on internal lapack, numpy's lapack-info routine should only be called when external lapack or site.cfg says so.

```bash
gcc -pthread -shared -L/om/user/satra/miniconda/lib -Wl,-rpath=/om/user/satra/miniconda/lib,--no-as-needed build/temp.linux-x86_64-2.7/nipy/labs/bindings/linalg.o -L/om/user/satra/miniconda/lib -L/om/user/satra/miniconda/lib -Lbuild/temp.linux-x86_64-2.7 -lcstat -lmkl_lapack95_lp64 -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lpython2.7 -o build/lib.linux-x86_64-2.7/nipy/labs/bindings/linalg.so
/usr/bin/ld: cannot find -lmkl_lapack95_lp64
collect2: error: ld returned 1 exit status
/usr/bin/ld: cannot find -lmkl_lapack95_lp64
collect2: error: ld returned 1 exit status
error: Command "gcc -pthread -shared -L/om/user/satra/miniconda/lib -Wl,-rpath=/om/user/satra/miniconda/lib,--no-as-needed build/temp.linux-x86_64-2.7/nipy/labs/bindings/linalg.o -L/om/user/satra/miniconda/lib -L/om/user/satra/miniconda/lib -Lbuild/temp.linux-x86_64-2.7 -lcstat -lmkl_lapack95_lp64 -lmkl_intel_lp64 -lmkl_intel_thread -lmkl_core -liomp5 -lpthread -lpython2.7 -o build/lib.linux-x86_64-2.7/nipy/labs/bindings/linalg.so" failed with exit status 1
```

alexis-roche:
My impression is that, whenever numpy's get lapack info is called, the external library linking options are passed on to gcc, resulting in a crash at compilation time. It may not be a problem of python not finding the libs, but rather one of passing undesired options to gcc. Am I correct?

matthew-brett:
Satra - are you explicitly asking to link against external lapack here, with the environment variable or nipy setup.cfg file?    As far as I can see, the lapack libraries and includes do not get added to the link step if you don't do that.  If you do do that, it seems reasonable that this will fail at the link step, if the libraries are not available.

satra:
@matthew-brett - i was using default master or this branch. both had the exact same effect. the default site.cfg was used.

```
[lapack]
# Value of 0 or False implies compile of, link to lapack_lite
# Value of 1 or True will cause setup to try and link to external BLAS /
# LAPACK as identified with the numpy configuration.  Default is False.
# The value in this file overrides the equivalent setting in the environment
# variable NIPY_EXTERNAL_LAPACK.
#external = False
```

satra:
these are all the places `get_info` is being called and that would have added those libraries.

```
./algorithms/clustering/setup.py:
   12 :     from numpy.distutils.system_info import get_info
   13 :     lapack_info = get_info('lapack_opt',0)
   17 :         lapack_info = get_info('lapack',0)
./labs/bindings/setup.py:
   12 :     from numpy.distutils.system_info import get_info
   13 :     lapack_info = get_info('lapack_opt',0)
   17 :         lapack_info = get_info('lapack',0)
./labs/glm/setup.py:
    8 :     from numpy.distutils.system_info import get_info
    9 :     lapack_info = get_info('lapack_opt', 0)
   13 :         lapack_info = get_info('lapack', 0)
./labs/group/setup.py:
    6 :     from numpy.distutils.system_info import get_info
    9 :     lapack_info = get_info('lapack_opt', 0)
   13 :         lapack_info = get_info('lapack',0)
./labs/setup.py:
   47 :     from numpy.distutils.system_info import get_info, system_info
   61 :     # XXX: We need to better sort out the use of get_info() for Lapack, because
   64 :     # get_info('lapack_opt') does not return the keys: 'libraries' and
   65 :     # 'library_dirs', but get_info('lapack') does.
   74 :     lapack_info = get_info('lapack_opt', 0)
   78 :         lapack_info = get_info('lapack', 0)
./labs/utils/setup.py:
    8 :     from numpy.distutils.system_info import get_info
    9 :     lapack_info = get_info('lapack_opt', 0)
   13 :         lapack_info = get_info('lapack', 0)
```

alexis-roche:
My memory is failing me, but reading again the discussion about https://github.com/nipy/nipy/pull/326, I see that I decided to remove any call to `get_info` to make it compile with MKL, which led to https://github.com/nipy/nipy/pull/328. Unfortunately, in the later PR, I also removed the option to compile via lapack_lite.

matthew-brett:
I'm just trying, with stripping all lapack references from the setup files (apart from the root nipy/labs one).  Compiles / runs OK on my OSX machine, trying now on travis : https://travis-ci.org/matthew-brett/nipy/builds/78460738

matthew-brett:
Those compiles also look OK, so I'm adding these edits to this PR.  

Log for non-external compile : https://s3.amazonaws.com/archive.travis-ci.org/jobs/78460742/log.txt

Log for external lapack compile : https://s3.amazonaws.com/archive.travis-ci.org/jobs/78460745/log.txt

Satra - does it work now?  If not, can you post the full log somewhere?

satra:
@matthew-brett - this branch works for me now even with MKL, with the default site.cfg and no environmental variable specification - thank you.

PR #355, opened 2015-09-02
BF: fix bug in writing PCA time-courses

matthew-brett:
We were using the axis number of the PCA analysis (usually axis index 3, time)
as the number of components (by default, 10).  Fix.

I've tested this at the command line with matplotlib, but it's hard to test
automatically without doing some image comparison test, and that seems like a
lot of work.


matthew-brett:
Last call for comments.

PR #356, opened 2015-09-02
MRG: refactor to common Python 2 / 3 codebase

matthew-brett:
Remove use of 2to3 and update to Python 3 idioms.

Remove some code for Python 2.5 compatibility.
Update code copies from latest upstream.
Remove doctest skipping machinery - a bit complicated for not much gain.


matthew-brett:
Any comments here?   I suggest reviewing by commit, rather than all changes at once, because there are a lot of changes here, and scrolling through lots of 'print(something)` changes can get pretty boring.

Commits of particular interest are 

* 33754ab - adding utilities module
* 99b36d8 - replacing some Python routines with utilities module functions
* 18574c9 - removing a mechanism to skip particular doctests (was a bit too fancy, causing problems).

nipy/algorithms/graph/graph.py
```diff
@@ -680,7 +680,7 @@ def dijkstra(self, seed=0):
         dist, active = np.inf * np.ones(self.V), np.ones(self.V)
         idx, neighb, weight = self.compact_neighb()
         dist[seed] = 0
-        dg = zip(np.zeros_like(seed), seed)
+        dg = list(zip(np.zeros_like(seed), seed))
```
yarikoptic: zip in PY2 returns a list, so listing a list is a waste.  you could  `from six.moves import zip`

nipy/algorithms/graph/tests/test_forest.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
```
yarikoptic: are you aiming to support 2.6? with 2.7 -- is this still necessary?

nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -478,7 +479,7 @@ def _getdtype(self):
     dtype = property(_getdtype, doc='The dtype of the design matrix of the Formula.')
 
     def __repr__(self):
-        return """Formula(%s)""" % `list(self.terms)`
+        return "Formula(%s)" % repr(list(self.terms))
```
yarikoptic: or just use `%r` and strip `repr()`?

nipy/algorithms/statistics/formula/formulae.py
```diff
@@ -526,7 +527,7 @@ def fromrec(rec, keep=[], drop=[]):
         if keep:
             return np.sum([t for n, t in f.items() if n in keep])
         else:
-            return np.sum(f.values())
+            return np.sum(list(f.values()))
```
yarikoptic: why explicit 'list' is needed here?

nipy/algorithms/statistics/models/glm.py
```diff
@@ -88,7 +90,7 @@ def fit(self, Y):
         self.scale = self.results.scale = self.estimate_scale()
 
         while self.cont():
-            self.results = self.next()
+            self.results = next(self)
```
yarikoptic: can't recall why I did import next from six in my code? ;-)

nipy/core/image/image.py
```diff
@@ -643,10 +645,10 @@ def rollaxis(img, axis, inverse=False):
         if type(axis) != type(0):
             raise ValueError('If carrying out inverse rolling, '
                              'axis must be an integer')
-        order = range(1, img.ndim)
+        order = list(range(1, img.ndim))
```
yarikoptic: again -- might be worth taking range from six.moves

nipy/core/image/tests/test_image.py
```diff
@@ -434,11 +434,11 @@ def test_rollimg_rollaxis():
     data = np.random.standard_normal((3,4,7,5))
     aff = np.diag([1,2,3,4,1])
     img = Image(data, AT('ijkl', 'xyzt', aff))
-    for axis in range(4) + range(-3, -1):
+    for axis in list(range(4)) + list(range(-3, -1)):
```
yarikoptic: ```for axis in itertools.chain(range(4), range(-3, -1)):``` ?

tools/nicythize
```diff
@@ -181,7 +181,7 @@ def find_process_files(root_dir):
     hash_db = load_hashes(HASH_FILE)
     for cur_dir, dirs, files in os.walk(root_dir):
         for filename in files:
-            for fromext, rule in rules.iteritems():
+            for fromext, rule in rules.items():
```
yarikoptic: there is six.iteritems, so would be `iteritems(rules)`

nipy/algorithms/graph/graph.py
matthew-brett: In terms of performance, listing a list is really trivial (a few microseconds).   The reason I like it is because, in the next few years where Python 3 takes over, it's clearer what is intended - more Python 3-ey and less six-ey.

nipy/algorithms/graph/tests/test_forest.py
matthew-brett: Are you asking if `from __future__ import print_function` is necessary?   Yes, I believe so...

nipy/algorithms/statistics/formula/formulae.py
matthew-brett: Good point - will do.

nipy/algorithms/statistics/formula/formulae.py
matthew-brett:
```
In [1]: a = {'a': 1, 'b': 2}

In [2]: import numpy as np

In [3]: np.sum(a.values())
Out[3]: dict_values([2, 1])
```

The list values are an iterator, and we need something that gets converted to an array.

nipy/algorithms/statistics/models/glm.py
matthew-brett: :) - if you find out - say?  Maybe Python 2.5 compatibility?

nipy/core/image/image.py
matthew-brett: Still prefer explicit Python 3 version - Python 2 cost very slight.

nipy/core/image/tests/test_image.py
matthew-brett: Nice - I've used that in a few places now.

tools/nicythize
matthew-brett: If we needed the iterator, that would be better, but I think list (Py 2) or iterator (Py 3) is OK here.

nipy/algorithms/graph/tests/test_forest.py
yarikoptic: On Thu, 03 Sep 2015, Matthew Brett wrote:

> In [1]nipy/algorithms/graph/tests/test_forest.py:

> > @@ -1,3 +1,5 @@
> > +from __future__ import print_function
> > +

> Are you asking if from __future__ import print_function is necessary? Yes, I
> believe so...

why?  

python2.7 -c 'print("from a function")'

works just fine.  Thanks for educating ;)

nipy/algorithms/statistics/formula/formulae.py
yarikoptic: > In [1]: a = {'a': 1, 'b': 2}

> In [2]: import numpy as np

> In [3]: np.sum(a.values())
> Out[3]: dict_values([2, 1])

> The list values are an iterator, and we need something that gets converted to
> an array.

ah... which python/numpy was that? with 2.7.10

    In [6]: a = {'a': 1, 'b': 2}

    In [7]: np.sum(a.values())
    Out[7]: 3

    In [8]: np.__version__
    Out[8]: '1.8.2'

that is why I have assumed that it just works and no explicit 'list'ing is
needed.  gotcha now

nipy/algorithms/graph/graph.py
yarikoptic: > > -        dg = zip(np.zeros_like(seed), seed)
> > +        dg = list(zip(np.zeros_like(seed), seed))

> In terms of performance, listing a list is really trivial (a few microseconds).
> The reason I like it is because, in the next few years where Python 3 takes
> over, it's clearer what is intended - more Python 3-ey and less six-ey.

gotcha, wise man!

nipy/algorithms/graph/tests/test_forest.py
matthew-brett: Yes, but that works in any Python version, and is printing a tuple:

```
$ python2.6 -c 'print("some", "strings")'
('some', 'strings')
$ python2.7 -c 'print("some", "strings")'
('some', 'strings')
$ python3.4 -c 'print("some", "strings")'
some strings
```

But maybe you meant - could I just print the tuple so it looks like it's a print function call?

nipy/algorithms/statistics/formula/formulae.py
matthew-brett: This a Python 3 thing - I get `dict_values([2, 1])` for numpy 1.8.2, on Python 3.

matthew-brett:
Any thoughts on the code removal?   The new utilities?

nipy/algorithms/graph/tests/test_forest.py
yarikoptic: On Fri, 04 Sep 2015, Matthew Brett wrote:

> In [1]nipy/algorithms/graph/tests/test_forest.py:

> > @@ -1,3 +1,5 @@
> > +from __future__ import print_function
> > +

> Yes, but that works in any Python version, and is printing a tuple:

> $ python2.6 -c 'print("some", "strings")'
> ('some', 'strings')
> $ python2.7 -c 'print("some", "strings")'
> ('some', 'strings')
> $ python3.4 -c 'print("some", "strings")'
> some strings

> But maybe you meant - could I just print the tuple so it looks like it's a
> print function call?

that was just awesome!  Thank you matthew for bearing with me ;)

matthew-brett:
Last call for comments - anyone?

PR #359, opened 2015-09-06
MRG: add mode and cval arguments for resampling

matthew-brett:
We sometimes need to resample images with specific ndimage resampling modes,
such as changing the constant values outside the boundary or the boundary
interpolation modes.

Add and test these arguments to the nipy resampling functions.


matthew-brett:
This is #309 with some style and docstring edits, a fix, and some extra tests.

matthew-brett:
@jbpoline - do you have time for a review?

matthew-brett:
@alexis-roche - when I was extending the tests, I noticed that the `resample` function in `algorithms/registration/resample.py` was broken for not-Affine transforms.  I added some fixes and more tests.  Would you mind taking a look?

alexis-roche:
I am proposing a patch to make the cython cubic spline resampling routine signature consistent with the simple sampling routine (`_cspline_sample`) used in the non-affine case. It did not make sense that the casting were handled at different levels depending on whether the transform was affine or not. The patch remedies this inconsistency:

https://github.com/alexis-roche/nipy/tree/matthew-resample

matthew-brett:
Do you also want to threshold the peak values so they cast correctly?  There's stuff in nibabel for that : 

```
from nibabel.casting import shared_range

mn, mx = shared_range(np.double, out_dtype)
res = np.clip(out, mn, mx).astype(out_dtype)
```

That ensures the top values don't wrap around into negatives etc.

matthew-brett:
Would you mind making your patch a PR against my branch?

alexis-roche:
Not sure users would expect clipping in the context of resampling an image, where the dynamic range is not expected to change significantly. I rather assume otherwise.

matthew-brett:
Probably the user wouldn't expect this though:

```
In [4]: a = np.array([0., 257.])

In [5]: a.astype(np.uint8)
Out[5]: array([0, 1], dtype=uint8)
```

alexis-roche:
Oh sorry - you mean the problem of wrapping around negative values. Yes, it would make sense to avoid that. But it should not affect intensities within the 'castable' range.

matthew-brett:
Right - anything outside the castable range with do terrible things with a naive cast.

matthew-brett:
Thanks - sorry to be slow to get back to this.

PR #360, opened 2015-09-20
MRG: more Python 3 fixes

matthew-brett:
Using an automated tool to check for Python 3 compatibility problems.


matthew-brett:
I know this one looks intimidating, but it's only very simple changes.   Should take 10 minutes or so to scroll through.  Any takers?

bthirion:
LGTM, no particular issue.

nipy/algorithms/statistics/onesample.py
```diff
@@ -132,6 +135,6 @@ def estimate_varatio(Y, sd, df=None, niter=10):
     value['random'] = sigma2
 
     if squeeze:
-        for key in value.keys():
+        for key in list(value):
             value[key] = np.squeeze(value[key])
```
yarikoptic: Why list is needed here?

matthew-brett: Hmm.  Python 3 doesn't allow you to modify the keys when iterating over the dictionary keys, but it does allow you to modify the values, so I think you're right, list isn't necessary.

```
In [11]: a = dict(one=1, two=2)

In [12]: for key in a:
   ....:     if key == 'two': a['three'] = 3
   ....:     
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-12-6b63c5f5ae33> in <module>()
----> 1 for key in a:
      2     if key == 'two': a['three'] = 3
      3 

RuntimeError: dictionary changed size during iteration

In [13]: for key in a:
    if key == 'two': a['one'] = 3
   ....:     

In [14]: 
```

nipy/algorithms/clustering/gmm.py
```diff
@@ -217,8 +218,8 @@ def plot2D(x, my_gmm, z=None, with_dots=True, log_scale=False, mpaxes=None,
         if z is None:
             plt.plot(x[:, 0], x[:, 1], 'o')
         else:
-            hsv = plt.cm.hsv(range(256))
-            col = hsv[range(0, 256, 256 // int(z.max() + 1))]
+            hsv = plt.cm.hsv(list(range(256)))
+            col = hsv[list(range(0, 256, 256 // int(z.max() + 1)))]
```
effigies: These two changes have no effect on the left-hand sides in Python >= 3.4. (Can't test lower, but don't see any reason from the docs that they should.)

nipy/algorithms/clustering/tests/test_gmm.py
```diff
@@ -110,7 +111,7 @@ def test_em_selection():
     dim = 2
     x = np.concatenate((nr.randn(100, dim), 3 + 2 * nr.randn(100, dim)))
 
-    krange = range(1, 10)
+    krange = list(range(1, 10))
```
effigies: Also unnecessary list construction. (`krange` iterated over in [best_fitting_GMM](https://github.com/matthew-brett/nipy/blob/more-py3-fixes/nipy/algorithms/clustering/gmm.py#L134).)

nipy/algorithms/clustering/tests/test_vmm.py
```diff
@@ -49,7 +50,7 @@ def test_dimension_selection_bic():
     x = (x.T / np.sqrt(np.sum(x**2, 1))).T
 
     precision = 100.
-    my_vmm = select_vmm(range(1,8), precision, False, x)
+    my_vmm = select_vmm(list(range(1,8)), precision, False, x)
```
effigies: Again.

nipy/algorithms/clustering/tests/test_vmm.py
```diff
@@ -64,7 +65,7 @@ def test_dimension_selection_cv():
 
     precision = 50.
     sub = np.repeat(np.arange(10), 2)
-    my_vmm = select_vmm_cv(range(1,8), precision, x, cv_index=sub,
+    my_vmm = select_vmm_cv(list(range(1,8)), precision, x, cv_index=sub,
```
effigies: Again.

nipy/algorithms/clustering/von_mises_fisher_mixture.py
```diff
@@ -415,7 +416,7 @@ def example_noisy():
     x = (x.T / np.sqrt(np.sum(x ** 2, 1))).T
 
     precision = 100.
-    vmm = select_vmm(range(2, 7), precision, True, x)
+    vmm = select_vmm(list(range(2, 7)), precision, True, x)
```
effigies: Again.

nipy/algorithms/clustering/von_mises_fisher_mixture.py
```diff
@@ -435,7 +436,7 @@ def example_cv_nonoise():
 
     precision = 50.
     sub = np.repeat(np.arange(10), 3)
-    vmm = select_vmm_cv(range(1, 8), precision, x, cv_index=sub,
+    vmm = select_vmm_cv(list(range(1, 8)), precision, x, cv_index=sub,
```
effigies: Again.

nipy/algorithms/clustering/gmm.py
matthew-brett: I guessed that was true, but I left these in because we otherwise depend on the fact that `plt.cm.hsv` accepts an iterator as well as a sequence, which I suppose is not guaranteed, but just happens to be the case (nothing I could see in the docstring to help).

nipy/algorithms/clustering/tests/test_gmm.py
matthew-brett: Same for these guys - in each case you're right - the list gets used inside the function as a iterable, but the docstring of the function says it should be a list.   Later on, I guess someone could change the implementation so that it does actually need a sequence, and then this would either break, or something quiet and horrible would happen.

The general idea is that - I would prefer to pass a list, unless an iterable is specifically stated to be acceptable.    Seem reasonable?

effigies: Seems fair. Will update my flagging criteria.

nipy/algorithms/clustering/gmm.py
effigies: Okay on `hsv`. What about indexing numpy arrays? There are some other ones where this also happens.

An alternative is to use slicing instead of an iterator:

    col = hsv[slice(0, 256, 256 // int(z.max() + 1))]

or

    col = hsv[::(256 // int(z.max() + 1))]

matthew-brett: For clarity I'd rather index numpy arrays with lists, but I don't feel crazy strongly.

The second slicing thing looks nicer to me - thanks.

matthew-brett: Sorry - I mean lists rather than range iterables.

effigies: Given that range iterables are indexed and reusable, I would recommend (in an off-topic way) dropping that bias, especially if the option is to call `list(range(...))` instead of `range(...)`. `int(x) in range(y)` [is now](https://docs.python.org/dev/library/stdtypes.html#typesseq-range) a O(1) lookup, as opposed to O(n) for `int(x) in list(range(y))`.

More relevant to the current PR, I think I would recommend against indexing numpy arrays by `range` and not `slice`, simply because a `range` is isomorphic to a `slice` (both are jumped-up `(start, stop, step)` tuples), and because there's a built-in slice notation. But perhaps this should be a new PR, since it's more style than notation?

nipy/algorithms/group/parcel_analysis.py
```diff
@@ -267,7 +268,7 @@ def __init__(self, con_imgs, parcel_img, parcel_info=None,
         # the input full-width-at-half-maximum parameter given in mm
         # to standard deviation in voxel units.
         orient = io_orientation(self.affine)[:, 0].astype(int)
-        voxsize = np.abs(self.affine[(orient, range(3))])
+        voxsize = np.abs(self.affine[(orient, list(range(3)))])
```
effigies: voxsize = np.abs(self.affine[orient, :3])

nipy/algorithms/segmentation/segmentation.py
```diff
@@ -265,5 +266,5 @@ def binarize_ppm(q):
     Assume input ppm is masked (ndim==2)
     """
     bin_q = np.zeros(q.shape)
-    bin_q[(range(q.shape[0]), np.argmax(q, axis=1))] = 1.
+    bin_q[(list(range(q.shape[0])), np.argmax(q, axis=1))] = 1.
```
effigies: Slice notation:

    bin_q[:q.shape[0], np.argmax(q, axis=1)] = 1.

nipy/algorithms/statistics/models/tests/test_model.py
```diff
@@ -15,7 +16,7 @@
 
 N = 10
 X = np.c_[np.linspace(-1,1,N), np.ones((N,))]
-Y = np.r_[range(5), range(1,6)]
+Y = np.r_[list(range(5)), list(range(1,6))]
```
effigies: From the docs: `np.r_` "Translates slice objects to concatenation along the first axis." Here's real slices:

    Y = np.r_[:5, 1:6]

nipy/core/image/image.py
```diff
@@ -220,7 +221,7 @@ def reordered_reference(self, order=None):
         )
         """
         if order is None:
-            order = range(self.ndim)[::-1]
+            order = list(range(self.ndim))[::-1]
```
effigies: Would drop.

    >>> list(range(5)[::-1]) == list(range(5))[::-1]
    True

nipy/core/reference/spaces.py
```diff
@@ -440,7 +441,7 @@ def xyz_order(coordsys, name2xyz=None):
             axvals[i] = N+i
         else:
             axvals[i] = 'xyz'.index(xyz_char)
-    if not set(axvals).issuperset(range(3)):
+    if not set(axvals).issuperset(list(range(3))):
```
effigies: Doesn't change anything with `issuperset`.

effigies:
LGTM. Couple notes about listification, particularly when slices could be used, but semantically see no reason things shouldn't work as is.

nipy/algorithms/clustering/gmm.py
matthew-brett: I agree, I would prefer to avoid indexing with `range`, because it's harder to see what numpy is going to do with a `range` in Python 3.  It looks like it does the right thing, but I don't know where that is guaranteed.

matthew-brett: Fixed hsv slicing as suggested.

nipy/algorithms/group/parcel_analysis.py
matthew-brett: Yup.

nipy/algorithms/segmentation/segmentation.py
matthew-brett: Much better.

nipy/algorithms/statistics/models/tests/test_model.py
matthew-brett: That is better, I agree.

nipy/core/reference/spaces.py
matthew-brett: Agree.

matthew-brett:
I rebased - think this is ready to merge now.

nipy/algorithms/group/parcel_analysis.py
```diff
@@ -267,7 +268,7 @@ def __init__(self, con_imgs, parcel_img, parcel_info=None,
         # the input full-width-at-half-maximum parameter given in mm
         # to standard deviation in voxel units.
         orient = io_orientation(self.affine)[:, 0].astype(int)
-        voxsize = np.abs(self.affine[(orient, range(3))])
+        voxsize = np.abs(self.affine[orient, :3])
```
matthew-brett: Oh dear - this change causes a change in the indexing, for reasons I am unable to understand or explain.  This version returns the 3 x 3 top corner of the affine, instead of a (3,) vector that is the intention.  Nasty.  Reverting.



http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#advanced-indexing

effigies:
Had a look at that indexing issue. I was working on the assumption that `orient` was a single integer, not a vector. Let `self.affine = np.eye(4)`. Then `orient = array([0, 1, 2])`.

    np.eye(4)[[0, 1, 2], :3] == np.eye(4)[[0, 1, 2]][:, :3] == np.eye(3)
    np.eye(4)[[0, 1, 2], range(3)] == np.eye(4)[[0, 1, 2], [0, 1, 2]] == np.ones(3)

May be good to add a comment, indicating `orient` is a (3, ) vector, so advanced indexing (link) is invoked?

nipy/algorithms/group/parcel_analysis.py
```diff
@@ -267,7 +268,8 @@ def __init__(self, con_imgs, parcel_img, parcel_info=None,
         # the input full-width-at-half-maximum parameter given in mm
         # to standard deviation in voxel units.
         orient = io_orientation(self.affine)[:, 0].astype(int)
-        voxsize = np.abs(self.affine[(orient, range(3))])
+        # `orient` is an array, so this slicing leads to advanced indexing.
```
matthew-brett: Added a comment.   I still don't understand the difference between the list and the slice here, but oh well.

effigies:
LGTM

matthew-brett:
Thanks for taking time to review, it was helpful.

PR #362, opened 2015-09-25
MRG: hack to work round np.intp bug in ndimage

matthew-brett:
Scipy ndimage, at least for scipy 0.9.0, fails when trying to interpolate data
of type np.intp, on a 32-bit system we are testing on:

http://nipy.bic.berkeley.edu/builders/nipy-py2.6-32/builds/291/steps/shell_6/logs/stdio

The interpolation failure gives an error like this:

```
File "/usr/lib/python2.7/dist-packages/scipy/ndimage/interpolation.py", line 392, in affine_transform
    output, order, mode, cval, None, None)
    RuntimeError: data type not supported
```

In fact, the numpy types np.int32 and np.intp (on 32-bits) are the same 32-bit
signed integer types, so we can just view the data as np.int32 (on 32 bits)
before running the ndimage routines.

Write wrappers around `affine_transform` and `map_coordinates` to do this
conversion, and extend tests.

The tests raise some errors on very old scipy, so update scipy dependency to
0.9.0.

### DIFF
--- a/nipy/fixes/scipy/ndimage/__init__.py
+++ b/nipy/fixes/scipy/ndimage/__init__.py
@@ -1,0 +1,87 @@
+""" Patches for scipy.ndimage
+
+Patched affine_transform to work round np.intp bug.
+
+Some versions of scipy.ndimage interpolation routines can't handle the np.intp
+type. I (MB) have only seen this on a 32-bit machine runing scipy 0.9.0
+"""
+
+import numpy as np
+
+import scipy.ndimage as spnd
+
+# int dtype corresponding to intp
+_INT_DTYPE = np.dtype(np.int32 if np.dtype(np.intp).itemsize == 4 else
+                      np.int64)
+
+
+# Do not run doctests on this module via nose; scipy doctests unreliable
+__test__ = False
+
+
+def _proc_array(array):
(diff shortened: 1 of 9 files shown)


matthew-brett:
Any comments here?

matthew-brett:
I'd be very glad of a review here, because it's a somewhat ugly hack.

nipy/fixes/scipy/ndimage/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+__test__ = False
+
+
+def _proc_array(array):
+    if array.dtype == np.dtype(np.intp):
+        return array.view(_INT_DTYPE)
+    return array
+
+
+def _proc_output(output):
```
bthirion: In case it matters, I find this function a bit suprising.
In particular
```
if np.dtype(output) == np.dtype(np.intp):
        return _INT_DTYPE
```
Since it returns a dtype ?
Maybe a short explanation is needed.

matthew-brett: I added some docstrings and comments - do they help?

bthirion: Sounds good, thx.

Issue #364, opened 2015-10-03
Data download site no longer works


matthew-brett:
Suggested site for data downloads in code and docs is http://nipy.org/data-packages - e.g. http://nipy.org/nipy/users/install_data.html

This no longer works, because we didn't port those over in the move from sourceforge.

They are still available via http://nipy.sourceforge.net/data-packages/

matthew-brett:
Probably fixed by the new https://github.com/nipy/data-packages gh-pages only repository.

PR #365, opened 2015-10-03
RF: move data package example to data package repo

matthew-brett:
We now have a dedicated data package repository to serve the data
packages.  Move the example data package there.

Probably closes gh-364.


bthirion:
That works for me. OK for merging.

matthew-brett:
Thanks for the review.

Issue #273, opened 2013-05-17
"slr" similarity in HistogramRegistration appears to be broken


matthew-brett:
In [29]: import nipy

    In [30]: from nipy.testing import funcfile, anatfile

    In [32]: anat_img = nipy.load_image(anatfile)

    In [33]: func_img = nipy.load_image(funcfile)

    In [34]: func_vol0 = func_img[..., 0]

    In [35]: register_obj = nar.HistogramRegistration(anat_img, func_vol0, similarity='slr')

    In [36]: register_obj.optimize()
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-36-2e999d7264a3> in <module>()
    ----> 1 register_obj.optimize()

    TypeError: optimize() takes at least 2 arguments (1 given)

    In [37]: register_obj.optimize('rigid')
    Initial guess...
    translation : [ 0.  0.  0.]
    rotation    : [ 0.  0.  0.]

    Optimizing using fmin_powell
    ---------------------------------------------------------------------------
    AttributeError                            Traceback (most recent call last)
    <ipython-input-37-9d9d4b937805> in <module>()
    ----> 1 register_obj.optimize('rigid')

    /home/mb312/usr/local/lib/python2.7/site-packages/nipy/algorithms/registration/histogram_registration.pyc in optimize(self, T, optimizer, **kwargs)
        282             print ('Optimizing using %s' % fmin.__name__)
        283         kwargs['callback'] = callback
    --> 284         Tv.param = fmin(cost, tc0, *args, **kwargs)
        285         return Tv.optimizable
        286 

    /usr/lib/python2.7/dist-packages/scipy/optimize/optimize.pyc in fmin_powell(func, x0, args, xtol, ftol, maxiter, maxfun, full_output, disp, retall, callback, direc)
    1608         direc = asarray(direc, dtype=float)
    1609 
    -> 1610     fval = squeeze(func(x))
    1611     x1 = x.copy()
    1612     iter = 0;

    /usr/lib/python2.7/dist-packages/scipy/optimize/optimize.pyc in function_wrapper(x)
        174     def function_wrapper(x):
        175         ncalls[0] += 1
    --> 176         return function(x, *args)
        177     return ncalls, function_wrapper
        178 

    /home/mb312/usr/local/lib/python2.7/site-packages/nipy/algorithms/registration/histogram_registration.pyc in cost(tc)
        251             # This is where the similarity function is calculcated
[...]

PR #367, opened 2015-10-07
add error checking for slr similarity measure in HistogramRegistration

alexis-roche:
This should fix https://github.com/nipy/nipy/issues/273

### DIFF
--- a/nipy/algorithms/registration/histogram_registration.py
+++ b/nipy/algorithms/registration/histogram_registration.py
@@ -189,7 +189,7 @@ class HistogramRegistration(object):
         if similarity in _sms:
             if similarity == 'slr':
                 if dist == None:
-                    raise ValueError('slr measure requires a joint intensity distribution model, '\
+                    raise ValueError('slr measure requires a joint intensity distribution model, '
                                      'see `dist` argument of HistogramRegistration')
                 if dist.shape != self._joint_hist.shape:
                     raise ValueError('Wrong shape for the `dist` argument')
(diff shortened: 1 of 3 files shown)


nipy/algorithms/registration/histogram_registration.py
```diff
@@ -180,6 +183,12 @@ def subsample(self, spacing=None, npoints=None):
 
     def _set_similarity(self, similarity, renormalize=False, dist=None):
         if similarity in _sms:
+            if similarity == 'slr':
+                if dist == None:
```
matthew-brett: Should be `if dist is None`

nipy/algorithms/registration/histogram_registration.py
```diff
@@ -180,6 +183,12 @@ def subsample(self, spacing=None, npoints=None):
...
+                    raise ValueError('slr measure requires a joint intensity distribution model, '\
```
matthew-brett: You don't actually need the slash at the end, because of the enclosing parens.

matthew-brett:
Thanks - that's kind of you.  Would you mind adding a test for the raised errors?

alexis-roche:
Yes, sorry, I was not expecting you to react so promptly and was actually preparing a test to be amended to the same commit.

alexis-roche:
Are we good here?

matthew-brett:
Yes, I can imagine you were expecting me to be slower :)

matthew-brett:
Thanks for doing that - looks fine - merging.

Issue #358, opened 2015-09-03
Intermittent memory error in `test_parcel_analysis_spm`


matthew-brett:
In latest travis test run - a memory error for one run in the matrix:

```
======================================================================
ERROR: nipy.algorithms.group.tests.test_parcel_analysis.test_parcel_analysis_spm
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/tests/test_parcel_analysis.py", line 80, in test_parcel_analysis_spm
    _test_parcel_analysis('spm', None)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/tests/test_parcel_analysis.py", line 53, in _test_parcel_analysis
    parcel_info=parcel_info)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/parcel_analysis.py", line 274, in __init__
    self._smooth_images(write_smoothed_images)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/parcel_analysis.py", line 291, in _smooth_images
    method=self.smooth_method)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/parcel_analysis.py", line 118, in _smooth_image_pair
    scon, svcon = smooth_fn(con, vcon, msk, sigma)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/parcel_analysis.py", line 92, in _smooth_spm
    K = _gaussian_energy(sigma)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.4/site-packages/nipy/algorithms/group/parcel_analysis.py", line 63, in _gaussian_energy
    x = np.zeros(mask_size)
MemoryError
```

See: https://travis-ci.org/matthew-brett/nipy/jobs/78490793

Testing of same commit a little earlier had given no memory error : https://travis-ci.org/matthew-brett/nipy/builds/78460738

Related to random numbers used in tests?

matthew-brett:
[...]

PR #368, opened 2015-10-08
Save memory in parcel analysis

alexis-roche:
This hopefully fixes https://github.com/nipy/nipy/issues/358

The way I had implemented the `_gaussian_energy` routine in parcel analysis was actually quite stupid as I was not exploiting kernel separability, and was therefore requesting quite a lot of memory, occasionally triggering memory errors.

### DIFF
--- a/nipy/algorithms/group/parcel_analysis.py
+++ b/nipy/algorithms/group/parcel_analysis.py
@@ -51,6 +51,20 @@ def _gaussian_filter(x, msk, sigma):
     return gx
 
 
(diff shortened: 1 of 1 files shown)


alexis-roche:
@matthew-brett Don't know if you can easily check that the random error doesn't occur anymore, I would be surprised if it did. The randomness didn't come from any random number generation, but simply from high memory overload.

matthew-brett:
Great, thanks for doing that.

PR #371, opened 2015-10-11
BF: fix a couple of examples

matthew-brett:
Python 3 adding of ranges in one example.

Stray text in another file causing a syntax error.


effigies:
Very straightforward. LGTM.

matthew-brett:
Chris - thanks for the review - merging.

Issue #369, opened 2015-10-11
empirical p-value error, maybe caused by numpy 1.10


matthew-brett:
This just came up on travis:

```
======================================================================
ERROR: nipy.algorithms.statistics.tests.test_empirical_pvalue.test_efdr
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/tests/test_empirical_pvalue.py", line 31, in test_efdr
    np.testing.assert_array_less(efdr.fdr(3.0), 0.2)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/empirical_pvalue.py", line 310, in fdr
    self.fdrcurve()
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/empirical_pvalue.py", line 234, in fdrcurve
    self.learn()
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/empirical_pvalue.py", line 206, in learn
    hist, ledge = np.histogram(x, bins=bins)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.5/site-packages/numpy/lib/function_base.py", line 249, in histogram
    n += np.bincount(indices, weights=tmp_w, minlength=bins).astype(ntype)
ValueError: operands could not be broadcast together with shapes (126,) (127,) (126,) 
----------------------------------------------------------------------
```
https://travis-ci.org/matthew-brett/nipy/jobs/84747582

I wonder whether this is an incompatibility with numpy 1.10 ?  I seem to remember numpy changed something subtle with its histogram bin selection in this release.

PR #372, opened 2015-10-13
BF: make histogram bins integer for numpy 1.10

matthew-brett:
Numpy 1.10 seems to need an integer input to np.histogram.

Probably fixes gh-369.

I would be happy of a quick review.

### DIFF
--- a/nipy/algorithms/statistics/empirical_pvalue.py
+++ b/nipy/algorithms/statistics/empirical_pvalue.py
@@ -202,7 +202,7 @@ class NormalEmpiricalNull(object):
 
         # generate the histogram
         step = 3.5 * np.std(self.x) / np.exp(np.log(self.n) / 3)
-        bins = max(10, (self.x.max() - self.x.min()) / step)
+        bins = max(10, (self.x.max() - self.x.min()) // step)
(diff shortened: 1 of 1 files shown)


effigies:
Looking at numpy/numpy#6469, it looks like it was rounding down before. Is this an explicit decision to round up? For consistent behavior across versions, I'd probably use integer division:

    bins = max(10, (self.x.max() - self.x.min()) // step)

To compare, another `np.histogram` call in this file [coerces to `int`](https://github.com/matthew-brett/nipy/blob/numpy-1.10-fix/nipy/algorithms/statistics/empirical_pvalue.py#L350). And another one looks like it [needs coercing/rounding](https://github.com/matthew-brett/nipy/blob/numpy-1.10-fix/nipy/algorithms/statistics/empirical_pvalue.py#L582). A little surprised that's not causing test failures.

matthew-brett:
Thanks for looking at this.  I made the // change you suggested.

The other np.histogram instance you found, I think is passing a vector of bin positions, so these do not have to be integers.

effigies:
Makes sense. CI error seems unrelated, so LGTM.

matthew-brett:
Thanks again for the review - in it goes.

Issue #370, opened 2015-10-11
ValueError in examples/labs/need_data/localizer_glm_ar.py


matthew-brett:
I think automated tests have found an error in this example.  Here's the log (line endings lost, sorry).  Bertrand - would you mind checking this example? 

```
Traceback (most recent call last): File "", line 1, in File "", line 141, in File "/Users/buildslave/osx-10.6/nipy-examples-3_4/build/venv/lib/python3.4/site-packages/nipy/modalities/fmri/glm.py", line 583, in contrast contrast_ = glm.contrast(con, contrast_type) File "/Users/buildslave/osx-10.6/nipy-examples-3_4/build/venv/lib/python3.4/site-packages/nipy/modalities/fmri/glm.py", line 236, in contrast resl = self.results_[l].Fcontrast(con_val) File "/Users/buildslave/osx-10.6/nipy-examples-3_4/build/venv/lib/python3.4/site-packages/nipy/algorithms/statistics/models/model.py", line 307, in Fcontrast matrix.shape[1])) ValueError: F contrasts should have shape[1] P=24, but this has shape[1] 25
```

http://nipy.bic.berkeley.edu/nipy-examples-3.4/labs-need_data-localizer_glm_ar.py.stderr

matthew-brett:
Closed by #373

PR #373, opened 2015-10-13
BF: fix design matrix size for example

matthew-brett:
I am not sure if this is the correct fix, but it's an attempted fix for
gh-370.

Bertrand - is this right?  But - why has the design matrix changed in
the number of columns?


matthew-brett:
Bertrand?   Can I merge this?  I think this is a 2 minute review.

bthirion:
LGTM

matthew-brett:
Thanks - merging.

PR #381, opened 2015-11-01
test_olsR: use assert_array_almost_equal instead

ignatenkobrain:
due to bug in numpy

Related: https://github.com/numpy/numpy/issues/5200

### DIFF
--- a/nipy/algorithms/statistics/models/tests/test_olsR.py
+++ b/nipy/algorithms/statistics/models/tests/test_olsR.py
@@ -148,9 +148,9 @@ def test_results():
     M = np.identity(14)
     M = np.array([M[i] for i in [0,1,2,3,4,5,6,8,9,10,11,12,13]])
     Fc = r.Fcontrast(M)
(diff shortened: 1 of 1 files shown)


bthirion:
Dear Igor,

     The fix sounds right, but beware that nipy contains lots of dead 
code and is actually no longer developed.
     Depending on your need, you might want to look at other projects.
     Best,

Bertrand Thirion

matthew-brett:
Great - thanks for doing this.  Looks good to me.

I will keep nipy in maintenance mode for at least another three years, probably five.

The code that you're testing here is I think already in `nistats` and will soon be in another package based on Jonathan Taylor's work for the current FMRI statistics stuff, and I can keep track of the correspondence for the second package - I guess Bertrand will be doing the same for `nistats`.

ignatenkobrain:
I'm packaging all projects related, so I'm working on packaging nistats also. There are problem with non compatible with new version of nilearn, but I hope this will be fixed soon.

bthirion:
It will, for sure. We were waiting for the next Nilearn release, to be compatible with release versions.

PR #384, opened 2015-11-29
Avoiding conditional directives that split up parts of statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/nipy/labs/group/glm_twolevel.c
+++ b/nipy/labs/group/glm_twolevel.c
@@ -7673,11 +7673,12 @@ static CYTHON_INLINE PyObject* __Pyx_PyNumber_Int(PyObject* x) {
   PyNumberMethods *m;
   const char *name = NULL;
   PyObject *res = NULL;
+  int checker;
+  checker = PyLong_Check(x);
 #if PY_MAJOR_VERSION < 3
-  if (PyInt_Check(x) || PyLong_Check(x))
-#else
-  if (PyLong_Check(x))
+  checker = PyInt_Check(x) || checker;
 #endif
+  if (checker)
     return __Pyx_NewRef(x);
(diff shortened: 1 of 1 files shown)


matthew-brett:
I see your point, but that particular file is auto-generated from Cython : http://cython.org - so edits will get wiped the next time we auto-generate from the `nipy/labs/group/glm_twolevel.pyx` file source.

You might consider sending a message to the Cython list about their auto-generated code?

matthew-brett:
Closing this one, please re-open if you disagree.

RomeroMalaquias:
Hello @matthew-brett  i'm from the same research team of flaviommedeirosand, i know that code is auto-generated, but i have to ask if you prefer this code before or after the commit. What do you think about this kind of change? 
Thanks in advance.

RomeroMalaquias:
I mean, would you accept if it was not for the fact we submitted the pull request to a code written by third parties?

matthew-brett:
The problem is that, this file is autogenerated from the Cython "pyx" file, so maintainability and readability of the C code are really not relevant - no-one should be editing this C code, because it will just get wiped out in the next rebuild of the Cython "pyx" files.

RomeroMalaquias:
@matthew-brett ,
oh i see, but If it was not a temporary code, would you accept it?
Sorry for insisting, I'm doing a research and your response is very valuable.

matthew-brett:
I'm not a C expert, but it looks reasonable to me.

PR #385, opened 2015-12-14
confusion variance <--> standard deviation

jokedurnez:
In the notes, variance and standard deviation were mixed.


bthirion:
Dear Joke,

The change is right of course -- many thanks. Please note however that the nipy package is no longer developed. Statistical analysis code should go to generic libraries (e.g. statsmodel) or more recent and  better-maintained libraries such as Nilearn or Nistats.

PR #390, opened 2016-01-04
BF: fix error message return for missing version

matthew-brett:
A bug in the setup_helpers where I was not returning the error
message when we can't get the version of a package we need.


matthew-brett:
I think this one's a simple fix - any chance of a review?

bthirion:
LGTM.

PR #393, opened 2016-03-12
BF: fix shebang lines for scripts

matthew-brett:
Shebang lines should be "!python" in order for setuptools / pip to
rewrite the shebangs when installing.

Fixes errors on Python 3.5 e.g.
http://nipy.bic.berkeley.edu/builders/nipy-bdist32-35/builds/6/steps/shell_10/logs/stdio

### DIFF
--- a/scripts/nipy_3dto4d
+++ b/scripts/nipy_3dto4d
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!python
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 DESCRIP = 'Read 3D image files and write a 4D file'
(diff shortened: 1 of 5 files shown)


yarikoptic:
correct description must be "break shebang lines for scripts" ;)
```
$> head /tmp/1.py
#!python
import sys
print("seems to be running")

$> /tmp/1.py
python: can't open file 'import sys
print("seems to be running")
': [Errno 2] No such file or directory
```
and here is description why shebangs should have full path: http://unix.stackexchange.com/a/77584

matthew-brett:
Yes, sure, this will mean that the scripts won't execute in place and have to be installed via distutils.  Distutils then writes in the correct full path.   Is there any need to have the full path in the scripts before installation?  If so, we'll have to do something complicated to make sure distutils does the right thing when installing into virtualenvs etc.

yarikoptic:
indeed, sounds reasonable ;)

yarikoptic:
also tested on precise in virtualenv -- seemed to do the right thing in both cases (pip install -e . and pip install .) so will just merge it on your behalf ;)

PR #394, opened 2016-03-12
MRG: drop threshold for scipy 0.17

matthew-brett:
It looks like scipy 0.17 linalg.lstsq (therefore pinv) has slightly different
precision than previous versions.  The differences in the pseudoinverse are
small, in the region of 1e-13 for a pseudoinverse of a matrix that is not very
close to singular.

See: https://mail.scipy.org/pipermail/scipy-dev/2016-March/021265.html

PR #397, opened 2016-03-19
MRG: prefer numpy pinv to scipy pinv

matthew-brett:
This is an alternative to #394.  In this PR, we switch from using
scipy.linalg.pinv to using numpy.linalg.pinv, on the basis that, at least as of
scipy 0.17, numpy pinv is more accurate, as measured by least squares sum of
squared errors.

This fixes the test failure we were getting from scipy 0.17 without the need to
relax the precision threshold (#394).

Of course, this PR might cause subtle changes in results.  On the other hand,
that will likely be true for scipy 0.17 compared to scipy 0.16 without this
change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ env:
         - INSTALL_TYPE="setup"
 python:
     - 2.6
-    - 3.2
     - 3.3
     - 3.4
     - 3.5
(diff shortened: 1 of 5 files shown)


matthew-brett:
I would be grateful for a review - this one will get our tests passing again.

bthirion:
LGTM.

matthew-brett:
Thanks for checking.

PR #400, opened 2016-05-06
ENH: pass kwargs to imshow also for plot_anat

mvdoc:
So that also anatomicals can be plotted with `interpolation='nearest'`, for example.

### DIFF
--- a/nipy/labs/viz_tools/activation_maps.py
+++ b/nipy/labs/viz_tools/activation_maps.py
@@ -235,7 +235,8 @@ def plot_map(map, affine, cut_coords=None, anat=None, anat_affine=None,
 
 
 def _plot_anat(slicer, anat, anat_affine, title=None,
-               annotate=True, draw_cross=True, dim=False, cmap=pl.cm.gray):
+               annotate=True, draw_cross=True, dim=False, cmap=pl.cm.gray,
+               **kwargs):
     """ Internal function used to plot anatomy
     """
     canonical_anat = False
@@ -274,7 +275,8 @@ def _plot_anat(slicer, anat, anat_affine, title=None,
             else:
                 vmin = vmean - (1+dim)*ptp
         slicer.plot_map(anat, anat_affine, cmap=cmap,
-                              vmin=vmin, vmax=vmax)
+                              vmin=vmin, vmax=vmax,
+                              **kwargs)
 
         if annotate:
             slicer.annotate()
@@ -296,7 +298,8 @@ def _plot_anat(slicer, anat, anat_affine, title=None,
 
 def plot_anat(anat=None, anat_affine=None, cut_coords=None, slicer='ortho',
               figure=None, axes=None, title=None, annotate=True,
-              draw_cross=True, black_bg=False, dim=False, cmap=pl.cm.gray):
+              draw_cross=True, black_bg=False, dim=False, cmap=pl.cm.gray,
+              **kwargs):
     """ Plot three cuts of an anatomical image (Frontal, Axial, and Lateral)
 
         Parameters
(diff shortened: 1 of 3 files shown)


yarikoptic:
If only there was a unittest... Could be mocked I guess to just verify that they are actually passed

mvdoc:
I should have expected it from you ;p added the unittest

mvdoc:
@yarikoptic mock is being installed by travis in this project

nipy/labs/viz_tools/test/test_activation_maps.py
```diff
@@ -3,6 +3,7 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 import tempfile
 
+from mock import patch
```
yarikoptic: awesome!  even if installed by travis, I don't see it used anywhere else :-/

try following
```python
try:
    from mock import patch
except ImportError:
    patch = None

```
then use patch not as a decorator but as a context handler within the test and decorate test with
```python
@skipif(not patch)
```
so the whole thing doesn't fail if mock is N/A

mvdoc:
@yarikoptic with Python 2.6 I cannot import `skipIf`. ideas?

yarikoptic:
use/import as in the other pieces of code here, e.g.
```shell
> git grep skipif
nipy/labs/spatial_models/tests/test_discrete_domain.py:from nipy.testing.decorators import skipif
nipy/labs/spatial_models/tests/test_discrete_domain.py:@skipif(not GOOD_GIFTI)
...
```

yarikoptic:
@matthew-brett any other desires I have not precluded? ;-)

nipy/labs/viz_tools/activation_maps.py
```diff
@@ -235,7 +235,8 @@ def plot_map(map, affine, cut_coords=None, anat=None, anat_affine=None,
 
 
 def _plot_anat(slicer, anat, anat_affine, title=None,
-               annotate=True, draw_cross=True, dim=False, cmap=pl.cm.gray):
+               annotate=True, draw_cross=True, dim=False, cmap=pl.cm.gray,
+               **kwargs):
```
matthew-brett: Maybe worth renaming to `plot_kwargs` for clarity?

matthew-brett:
Otherwise looks fine to me - Yarik - do you want to merge when you're happy?

mvdoc:
@yarikoptic @matthew-brett I renamed `kwargs` to `imshow_kwargs` for both `plot_anat` and `plot_map` for consistency

mvdoc:
@yarikoptic anything else?

yarikoptic:
I hate when decreased coverage forbids me merging

yarikoptic:
And bloody coveralls fails to show what is not covered. I guess it is due to mock, why don't you install it in Travis?

@matthew-bret I got to like codecov especially when its browser plugin works - just an ideal

yarikoptic:
We should enable codecov here... I will do it later

nipy/labs/viz_tools/test/test_activation_maps.py
```diff
@@ -15,6 +16,11 @@
 except ImportError:
     raise SkipTest('Could not import matplotlib')
 
+try:
+    from mock import patch
+except ImportError:
```
yarikoptic: ah, right -- this would be two new uncovered lines if mock is available.  Just add to `except` line above an inline comment `# pragma: no cover` which would then instruct coverage to exclude next block from its consideration.  Should bring coverage back (and show enabled codecov in action)

matthew-brett:
Oops - sorry - forgot about this one.  Looks fine to me - merging.

PR #401, opened 2016-05-06
DOC: add missing docstring for dim in plot_anat

mvdoc:


nipy/labs/viz_tools/activation_maps.py
```diff
@@ -342,6 +342,8 @@ def plot_anat(anat=None, anat_affine=None, cut_coords=None, slicer='ortho',
             you whish to save figures with a black background, you
             will need to pass "facecolor='k', edgecolor='k'" to pylab's
             savefig.
+        dim: float, optional
+            If set, dim the anatomical image.
```
yarikoptic: So is it opacity/alpha with 1 being in full, or amount of dimness ie 1-that?

mvdoc: no, I believe it selects the maximum value for plotting to be sth from the mean above
```
vmax = vmean + (1+dim)*ptp
```

I didn't know how to phrase it precisely

mvdoc:
ping @yarikoptic: better?

yarikoptic:
Better than it was ;-)

Issue #404, opened 2016-06-25
Intermittent doctest failure in mixed_effects_stat


matthew-brett:
```
======================================================================
FAIL: Doctest: nipy.algorithms.statistics.mixed_effects_stat
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.5/doctest.py", line 2189, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for nipy.algorithms.statistics.mixed_effects_stat
  File "/home/buildslave/bongoslave/nipy-py3_5-32/build/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/mixed_effects_stat.py", line 0, in mixed_effects_stat

----------------------------------------------------------------------
File "/home/buildslave/bongoslave/nipy-py3_5-32/build/venv/lib/python3.5/site-packages/nipy/algorithms/statistics/mixed_effects_stat.py", line 20, in nipy.algorithms.statistics.mixed_effects_stat
Failed example:
    assert np.array([t1.std() < t2.std() for t1, t2 in zip(T1, T2)]).all()
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib/python3.5/doctest.py", line 1320, in __run
        compileflags, 1), test.globs)
      File "<doctest nipy.algorithms.statistics.mixed_effects_stat[7]>", line 1, in <module>
        assert np.array([t1.std() < t2.std() for t1, t2 in zip(T1, T2)]).all()
    AssertionError
```

Examples:

* https://travis-ci.org/MacPython/nipy-wheels/jobs/140268248#L713
* http://nipy.bic.berkeley.edu/builders/nipy-py3.5-32/builds/1/steps/shell_7/logs/stdio

@bthirion - would you mind taking a look?

matthew-brett:
Sorry - investigating more - this failure seems to occur about 1 in 20 runs of the doctest, regardless of platform - so I suppose it is due to differences in the random numbers used.   Is it reasonable that this test can sometimes fail?

bthirion:
I haven't been able to find time for this -- I'm at OHBM currently. Will take a look whenever I find time together with a wifi connection.

matthew-brett:
Closed by #406

PR #406, opened 2016-08-05
[MRG] fixed failing doctest

bthirion:
Addresses #404


matthew-brett:
Sorry for my ignorance, but can you explain the change?

bthirion:
I removed the cases where the simulated effects are 0: In these cases, due to the constraint in l.376, the mfx test can be more variable then the classical Student alternative, leading to a failure of the docstring.

matthew-brett:
Thanks - that's helpful.

PR #408, opened 2016-09-04
BF: allow for generic IOError-based exception to be thrown when loading non existing file

yarikoptic:
NB not sure how this one escaped my testing of dependents when checking one of nibabel rcs

need it now to resole FTBFS on debians https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=836538


matthew-brett:
Thanks - sorry for the delay.

Issue #409, opened 2016-09-08
SpaceTimeRealign/Realign4d TR


TheChymera:
I noticed that SpaceTimeRealign/Realign4d really *really* want to be given a tr value explicitly, and if none is given [they fail](https://github.com/nipy/nipy/blob/master/nipy/algorithms/registration/groupwise_registration.py#L742).

I wonder why.
TR is contained in the NIfTI header.
I'm not necessarily protesting that the user has the choice to override (rather than fix) a bad header - but shouldn't at least the header be used if the user specifies no value?

I can write a fix for this if you would pull.

matthew-brett:
The problem is that the TR is often not contained in the NIfTI header.  It might be a default value like 0 in which case that's obvious, or a default value like 1, in which case it's not obvious.  So we'd need to decide some policy about when the TR value in the header is intentional vs accidental.

TheChymera:
I'd say unless the header value will lead to an inevitable exception further down the line (which I assume is the case for 0) taking it will be better than raising an error.

So I would propose the following policy:

* prefer the user-specified value, if None read from header
* raise an exception if the header value is 0

I don't think it's a good "feature" to make a function fail when it doesn't have to just because we suspect some users might not be entirely familiar with their data.
If anything we should print a warning, but I doubt that the people who are unaware of whether or not their headers are broken will be reading through the nipype output.

Not least of all - does Realign4d only work with 4d data? whether or not it accepst multiple dimensions will determine how best to read TR from the header.

matthew-brett:
The key thing for me is that it's not at all easy to get an answer you didn't intend.  Clearly 0 is wrong, and we can raise an error.   On the other hand, it's not clear whether 1 is wrong or not.  It could be the correct TR or it could be a default written into the header by software that didn't know the TR, or (if I remember right) the default value filled in by Nipy.   It would be very easy not to notice this and get a result you didn't intend.

TheChymera:
I don't think we should operate under the assumption that some value is invalid, just because people might not notice mistakes. If the value works, I think the reasonable assumption is that it is for real. I, for one, mainly use TR=1

Also, from what I have seen, if people are dealing with malformed headers, they always look to define TR explicitly. If they don't look to do that, it means they trust in their headers.

alexis-roche:
Hi Christian,

This brings us back to a discussion we had a while ago. We did consider the
policy you are suggesting, however it is a fact that the tr filled in the
nifti header cannot always be trusted and we thought it would just make
[...]

PR #410, opened 2016-09-09
Realign4d TR from header

TheChymera:
with "fail-safe" caveats. addressing https://github.com/nipy/nipy/issues/409

### DIFF
--- a/nipy/algorithms/registration/groupwise_registration.py
+++ b/nipy/algorithms/registration/groupwise_registration.py
@@ -24,14 +24,14 @@ from ...externals.six import string_types
 from nibabel.affines import apply_affine
 
 from ...fixes.nibabel import io_orientation
-
+from ...io.nibcompat import get_header
 from ...core.image.image_spaces import (make_xyz_image,
                                         xyz_affine,
                                         as_xyz_image)
 from ..slicetiming import timefuncs
-from .type_check import (check_type, check_type_and_shape)
-from .optimizer import configure_optimizer, use_derivatives
 from .affine import Rigid, Affine
+from .optimizer import configure_optimizer, use_derivatives
+from .type_check import (check_type, check_type_and_shape)
 from ._registration import (_cspline_transform,
                             _cspline_sample3d,
                             _cspline_sample4d)
@@ -86,6 +86,35 @@ def guess_slice_axis_and_direction(slice_info, affine):
         slice_direction = int(slice_info[1])
     return slice_axis, slice_direction
 
+def tr_from_header(images):
+    """ Return the TR from the header of an image or list of images.
+
+    Parameters
+    ----------
+    images : image or list of images
+      Single or multiple input 4d images representing one or
+      several sessions.
+
+    Returns
+    -------
+    float
+      Repetition time, as specified in NIfTI header.
+
+    Raises
+    ------
+    ValueError
+      if the TR between the images is inconsistent.
+    """
+    if not isinstance(images, list):
+        images = [images]
+    images_tr = None
+    for image in images:
+        tr = get_header(image).get_zooms()[3]
+        if images_tr is None:
+            images_tr = tr
+        if tr != images_tr:
+            raise ValueError('TR inconsistent between images.')
+    return images_tr
 
 class Image4d(object):
     """
@@ -736,14 +765,16 @@ class Realign4d(object):
             time_interp = False
         else:
             time_interp = True
-        self.slice_times = slice_times
-        self.tr = tr
-        if tr is None:
-            raise ValueError('Repetition time cannot be None')
         if not isinstance(images, (list, tuple, np.ndarray)):
             images = [images]
-        self._runs = []
+        if tr is None:
+            raise ValueError('Repetition time cannot be None.')
+        if tr == 0:
+            raise ValueError('Repetition time cannot be zero.')
         self.affine_class = affine_class
+        self.slice_times = slice_times
+        self.tr = tr
+        self._runs = []
         # Note that, the affine of each run may be different. This is
         # the case, for instance, if the subject exits the scanner
         # inbetween sessions.
@@ -839,7 +870,7 @@ class Realign4d(object):
             and 'steepest'.
         maxiter : int
             Maximum number of iterations in optimization.
-        maxfun : int 
+        maxfun : int
             Maximum number of function evaluations in maxfun.
         """
         if between_loops is None:
@@ -897,9 +928,11 @@ class SpaceTimeRealign(Realign4d):
         images : image or list of images
             Single or multiple input 4d images representing one or several fMRI
             runs.
-        tr : float
+        tr : None or float or "header-allow-1.0"
             Inter-scan repetition time in seconds, i.e. the time elapsed between
-            two consecutive scans.
+            two consecutive scans. If None, an attempt is made to read the TR
+            from the header, but an exception is thrown for values 0 or 1. A
+            value of "header-allow-1.0" will signal to accept a header TR of 1.
         slice_times : str or callable or array-like
             If str, one of the function names in ``SLICETIME_FUNCTIONS``
             dictionary from :mod:`nipy.algorithms.slicetiming.timefuncs`.  If
@@ -922,6 +955,17 @@ class SpaceTimeRealign(Realign4d):
             transformation class to use to calculate transformations between
             the volumes. Default is :class:``Rigid``
         """
+        if tr is None:
+            tr = tr_from_header(images)
+            if tr == 1:
+                raise ValueError('A TR of 1 was found in the header. '
+                    'This value often stands in for an unknown TR. '
+                    'Please specify TR explicitly. Alternatively '
+                    'consider setting TR to "header-allow-1.0".')
+        elif tr == "header-allow-1.0":
+            tr = tr_from_header(images)
+        if tr == 0:
+            raise ValueError('Repetition time cannot be zero.')
         if slice_times is None:
             raise ValueError("slice_times must be set for space/time "
                              "registration; use SpaceRealign for space-only "
--- a/nipy/algorithms/registration/tests/test_fmri_realign4d.py
+++ b/nipy/algorithms/registration/tests/test_fmri_realign4d.py
@@ -4,25 +4,29 @@ from __future__ import absolute_import
 
 import warnings
 
-from nose.tools import assert_equal
+import numpy as np
+import nibabel as nib
 
+from nose.tools import assert_equal
 from numpy.testing import (assert_array_almost_equal,
                            assert_array_equal,
                            assert_raises)
-import numpy as np
 
 from .... import load_image
-from ....testing import funcfile
-from ....fixes.nibabel import io_orientation
 from ....core.image.image_spaces import (make_xyz_image, xyz_affine)
-
+from ....fixes.nibabel import io_orientation
+from ....io.nibcompat import get_header
+from ....testing import funcfile
+from ...slicetiming.timefuncs import st_43210, st_02413, st_42031
+from ..affine import Rigid
 from ..groupwise_registration import (Image4d, resample4d, FmriRealign4d,
                                       SpaceTimeRealign, SpaceRealign, Realign4d,
                                       Realign4dAlgorithm, make_grid)
-from ...slicetiming.timefuncs import st_43210, st_02413, st_42031
-from ..affine import Rigid
 
 im = load_image(funcfile)
+ims = [nib.Nifti1Image(np.zeros((2, 3, 4, 5)), np.eye(4)) for i in range(4)]
+for ix, imx in enumerate(ims):
+    get_header(imx)['pixdim'][4] = ix
 
 def test_futurewarning():
     with warnings.catch_warnings(record=True) as warns:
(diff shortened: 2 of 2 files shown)


TheChymera:
@matthew-brett ok, so the logic has an issue: If the TR values are non-consistent, `tr_from_header` will return `None`, after the initial `tr is None` check.

Checking for None twice seems really clunky. I'd prefer to just raise an exception in `tr_from_header` - or do you see any purpose in it returning `None`?

matthew-brett:
It will be more obvious to you, doing the implementation, what the cleanest way is - please feel free to chose what you prefer.

TheChymera:
@matthew-brett It seems the checks are indeed failing because of something with the functions I've been working on. Can you help me figure out why? I don't understand what this [not raising a `ValueError`](https://travis-ci.org/nipy/nipy/jobs/159436030#L7920) is all about

matthew-brett:
What happens when you try running the same checks in IPython, like:

```
>>> SpaceTimeRealign(runs, None, [0, 1, 2], 2)
```

Do they generate a ValueError?  Should they?

TheChymera:
@matthew-brett It generates a NameError, as I believe it should, since `runs` is not defined.

matthew-brett:
Sure, but what if you replicate the environment in the test.  I guess `runs` is defined there?

TheChymera:
Hmmm, that gives me the output I would expect, but I am surprised that it's the last line of the exception that gets highlighted:

```
/home/chymera/src/nipy/nipy/algorithms/registration/groupwise_registration.py in _init(self, images, tr, slice_times, slice_info, affine_class)
    758                 raise ValueError('A TR of 1 was found in the header.\
    759                     This value often stands in for an unknown TR. Please\
--> 760                     specify TR explicitly.')
    761         elif tr is "header-allow-1.0":
    762             tr = tr_from_header(images)

ValueError: A TR of 1 was found in the header.                    This value often stands in for an unknown TR. Please                    specify TR explicitly.
```

Also, I don't like how that error message prints, so I just updated the PR, now it gives me this, which I guess is just as wrong/right, but looks a deal better:

```
In [17]: SpaceTimeRealign(img, None, [0, 1, 2], 2)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-c283a7595358> in <module>()
----> 1 SpaceTimeRealign(img, None, [0, 1, 2], 2)

/home/chymera/src/nipy/nipy/algorithms/registration/groupwise_registration.py in __init__(self, images, tr, slice_times, slice_info, affine_class)
    967         if hasattr(slice_times, '__call__'):
    968             slice_times = slice_times(n_slices, tr)
--> 969         self._init(images, tr, slice_times, slice_info, affine_class)
    970 
    971 

/home/chymera/src/nipy/nipy/algorithms/registration/groupwise_registration.py in _init(self, images, tr, slice_times, slice_info, affine_class)
    755             tr = tr_from_header(images)
    756             if tr == 1:
--> 757                 raise ValueError('A TR of 1 was found in the header. '\
    758                     'This value often stands in for an unknown TR. '\
    759                     'Please specify TR explicitly.')

ValueError: A TR of 1 was found in the header. This value often stands in for an unknown TR. Please specify TR explicitly.
```

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -737,9 +751,17 @@ def _init(self, images, tr, slice_times, slice_info, affine_class):
         else:
             time_interp = True
         self.slice_times = slice_times
-        self.tr = tr
         if tr is None:
-            raise ValueError('Repetition time cannot be None')
+            tr = tr_from_header(images)
+            if tr == 1:
+                raise ValueError('A TR of 1 was found in the header. '\
```
matthew-brett: You don't need the line continuation here, Python will combine the strings without them.

matthew-brett:
Sorry - I see - you are seeing the errors for the tests that were expecting None to generate an error.   You need to replace those tests with tests of the new TR-finding machinery.

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -86,6 +86,20 @@ def guess_slice_axis_and_direction(slice_info, affine):
...
+        no_init_value = True
+        for image in images:
+            tr = image.header.get_zooms()[-1]
+            if no_init_value:
+                images_tr = tr
+                no_init_value = False
+            if tr != images_tr:
+                raise ValueError('TR inconsistent between images.')
+        else:
+            return images_tr
```
matthew-brett: The `else` is a bit confusing here, you usually use this to run something when there was no `break` in the for loop.  Of course there isn't here, but an experienced programmer will look for one.  How about something like:

```
def tr_from_header(images):
    """ A good docstring """
    if not isinstance(images, list):
        images = [images]
    images_tr = None
    for image in images:
        tr = image.header.get_zooms()[3]
        if images_tr is None:
            images_tr = tr
        if tr != images_tr:
            raise ValueError('TR inconsistent between images.')
    return images_tr
```

TheChymera:
Currently we only get a ValueError if the header TR is 0 or 1 - which is not the case for `functional.nii.gz`. should I just remove the offending lines?

matthew-brett:
Why not write a few tests for the states that do raise errors, such as bad TRs or TR not equal between images?  You could make some example `runs` e.g.

```
import nibabel as nib
img = nib.Nifti1Image(np.zeros((2, 3, 4, 5), np.eye(4))
img.header['pixdim'][4] = 0
```

etc.

nipy/algorithms/registration/groupwise_registration.py
TheChymera: I went for `no_init_value` instead of the more elegant 

```
if images_tr is None:
            images_tr = tr
```

BEcause I was worried that the logic would be unable to pick up different-TR images if the first n of the list had `image.header.get_zooms()[3] == None` but it seems `.get_zooms()` can never produce `None`. Can you confirm this?

matthew-brett: Yes, that's right - it only produces a vector of numbers, one per dimension of the image.

TheChymera:
I can't really make sense of the coveralls output - how exactly did I reduce the coverage?

@matthew-brett  Also, can you help me out with travis? I don't get this:

```
AttributeError: 'Nifti1Image' object has no attribute 'header'
```

Afaict  [`Nifti1Image` objects *do* have a header](http://nipy.org/nibabel/nibabel_images.html#the-image-object).

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -737,9 +756,17 @@ def _init(self, images, tr, slice_times, slice_info, affine_class):
...
             time_interp = True
         self.slice_times = slice_times
-        self.tr = tr
         if tr is None:
-            raise ValueError('Repetition time cannot be None')
+            tr = tr_from_header(images)
+            if tr == 1:
+                raise ValueError('A TR of 1 was found in the header. '
+                    'This value often stands in for an unknown TR. '
+                    'Please specify TR explicitly.')
```
matthew-brett: Maybe add: `or consider "header-allow-1.0"`.

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -194,7 +198,9 @@ def test_spacetimerealign_params():
     R = SpaceTimeRealign(runs, 3, 'ascending', 2) # OK
     assert_raises(ValueError, SpaceTimeRealign, runs, 3, None, 2)
     assert_raises(ValueError, SpaceTimeRealign, runs, 3, 'ascending', None)
-    assert_raises(ValueError, SpaceTimeRealign, runs, None, [0, 1, 2], 2)
+    assert_raises(ValueError, SpaceTimeRealign, ims[0], None, [0, 1, 2], 2)
+    assert_raises(ValueError, SpaceTimeRealign, ims[1], None, [0, 1, 2], 2)
+    assert_raises(ValueError, SpaceTimeRealign, ims[2:4], None, [0, 1, 2], 2)
```
matthew-brett: Add a test with `header-allow-1.0` as well?

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -23,6 +24,9 @@
 from ..affine import Rigid
 
 im = load_image(funcfile)
+ims = [nib.Nifti1Image(np.zeros((2, 3, 4, 5)), np.eye(4)) for i in range(4)]
+for ix, imx in enumerate(ims):
```
matthew-brett: The problem is likely with nipy checking compatibility with old nibabel.  Try:

```
from nipy.io.nibcompat import get_header
...

    header = get_header(imx)
```

matthew-brett:
Sorry, the coveralls output is hard to read.   I now prefer `codecov`, I should set that up.   But I'm sure the problem is that you aren't currently testing the `header-allow-1.0` branch.  Try testing that option?

TheChymera:
@matthew-brett ok, it seems there is yet another problem. TR is needed in `SpaceTimeRealign` [before it is initialized](https://github.com/nipy/nipy/blob/master/nipy/algorithms/registration/groupwise_registration.py#L946).

What would you say is better - moving the `slice_times` calculation to `Realign4d`, or the entire TR logic to `SpaceTimeRealign`?

matthew-brett:
I'd say move that logic into the `_init` method, and modify the docstring for Realign4d to reflect that.

TheChymera:
I added the logic to `SpaceTimeRealign` instead, as there was no way to separate the tr from the rest of the higher-level logic. Also, modifying the input parsing for `Realign4d` might have broken other classes based on it. I think it makes sense to keep `Realign4d` as basic as possible and do the fancy stuff in other classes.

I still don't understand what's up with codecov. The lines it highlights are tested by [line 202 and 204, respectively](https://github.com/TheChymera/nipy/blob/5aeb294f626463ad67911982476bd37d160e0da2/nipy/algorithms/registration/tests/test_fmri_realign4d.py#L202).

matthew-brett:
Not sure what's going on with the coverage.  Have you tested that 'header-allow-1.0' works on an image with TR=0?

TheChymera:
@matthew-brett anything else you'd like to recommend before I squash this?

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -18,20 +18,18 @@
 import warnings
 
 import numpy as np
-
-from ...externals.six import string_types
-
 from nibabel.affines import apply_affine
 
+from ...externals.six import string_types
```
matthew-brett: This import should go at the top, because it's a sort-of special import, like a system import.

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -86,6 +84,25 @@ def guess_slice_axis_and_direction(slice_info, affine):
...
     return slice_axis, slice_direction
 
+def tr_from_header(images):
+    """ Return the TR from the header of an image or list of images.
+
+    Parameters
+    ----------
+    images : image or list of images
+      Single or multiple input 4d images representing one or
+      several sessions.
```
matthew-brett: Add `Returns` section.

matthew-brett: Add `Raises` section : https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

nipy/algorithms/registration/groupwise_registration.py
```diff
@@ -897,9 +914,11 @@ def __init__(self, images, tr, slice_times, slice_info,
         images : image or list of images
             Single or multiple input 4d images representing one or several fMRI
             runs.
-        tr : float
+        tr : None or float or string
```
matthew-brett: I'd prefer `None or float or 'header-allow-1.0'` to make the possible strings specific.

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -3,14 +3,16 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 
 import warnings
+import nibabel as nib
+import numpy as np
```
matthew-brett: Import numpy before nibabel (system imports first, then common imports, then package-specific imports).

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -23,6 +25,9 @@
 from ..affine import Rigid
 
 im = load_image(funcfile)
```
matthew-brett: Sorry - this isn't fixing what you did, but what was there before.  Would you mind capitalizing these to `IMG` and `IMAGES` to make it clear they are module globals.

TheChymera: I'll get on the rest, but I think it is best if we do this in a separate commit.

nipy/algorithms/registration/tests/test_fmri_realign4d.py
```diff
@@ -166,8 +171,8 @@ def test_realign4d_params():
     # Some tests for input parameters to realign4d
     R = Realign4d(im, 3, [0, 1, 2], None) # No slice_info - OK
     assert_equal(R.tr, 3)
-    # TR cannot be None for set slice times
-    assert_raises(ValueError, Realign4d, im, None, [0, 1, 2], None)
+    # TR cannot be zero
+    assert_raises(ValueError, Realign4d, ims[0], None, [0, 1, 2], None)
```
matthew-brett: Add check that passed TR (not from header) cannot be zero?

nipy/algorithms/registration/tests/test_fmri_realign4d.py
matthew-brett: Add test for error with `header-allow-1.0` and image TR=0?

nipy/algorithms/registration/groupwise_registration.py
matthew-brett: Nice, thanks.

nipy/algorithms/registration/tests/test_fmri_realign4d.py
matthew-brett: Add `assert_raises(ValueError, SpaceTimeRealign, ims[0], 'header-allow-1.0', [0, 1, 2], 2)`

matthew-brett:
Great - thanks a lot for your patience.

Issue #411, opened 2016-09-14
realign test code style


TheChymera:
@matthew-brett you [pointed out that some capitalization might be needed](https://github.com/nipy/nipy/pull/410#r78789690). What exactly should be capitalized - and per what recommendation? [This](https://www.python.org/dev/peps/pep-0008/#global-variable-names) is the only style guide I found for capitalization, and I believe it makes no recommendations pertaining to our image object instances.

matthew-brett:
I was suggesting using all-caps for the global variable names. No, it's not in the PEP, just a convention that is fairly common, and that I find helpful:

http://stackoverflow.com/questions/2601269/arguments-local-variables-and-global-variables-coding-convention-in-python

TheChymera:
From the answers in that thread you linked to - does it not seem more common that using different naming schemes for local and global variables is bad practice?

I for one had no trouble finding where the image object/lists were defined.

matthew-brett:
From that thread I see this "Why is there a need to distinguish between arguments and local variables" - but that's not relevant to us.  Then further down there's this:

"It is usually obvious in python which variables are local and which are global, since to modify a global variable you have to declare it using the the global keyword at the start of a function. However I sometimes add a global declaration even if it is not necessary for python to compile it, in order to emphasize that an object is global - e.g. modifying a mutable global data-structure.

Arguments should be obvious because they are in the function declaration.

As others have said constants should be in UPPER_CASE_WITH_UNDERSCORES, which is a convention shared by many languages."

So maybe I should have said it differently - I like module level global constants to be in uppercase, as this person does.  Module level global variables, used as variables rather than constants, will be easy to see, because they need `global` declarations inside the function using them.

TheChymera:
Ah, so im and ims [here](https://github.com/nipy/nipy/blob/master/nipy/algorithms/registration/tests/test_fmri_realign4d.py#L26) are constants, and they and only they should be capitalized - y?

matthew-brett:
Right.

PR #412, opened 2016-09-15
Capitalized constants

TheChymera:
Addresses #411

### DIFF
--- a/nipy/algorithms/registration/tests/test_fmri_realign4d.py
+++ b/nipy/algorithms/registration/tests/test_fmri_realign4d.py
@@ -23,27 +23,27 @@ from ..groupwise_registration import (Image4d, resample4d, FmriRealign4d,
                                       SpaceTimeRealign, SpaceRealign, Realign4d,
                                       Realign4dAlgorithm, make_grid)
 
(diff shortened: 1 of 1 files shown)


matthew-brett:
Excellent - thanks for doing that.

matthew-brett:
@nibotmi r+

TheChymera:
@matthew-brett one unrelated question: I header nipy is being discontinued? Is that true? What happens to all the nice native-python preprocessing functions when it does?

matthew-brett:
The general plan is to keep nipy maintained for the next five years or so, but more active development of various parts has shifted to: 

1) registration : https://github.com/nipy/nireg (although it looks like development is a bit slow);
2) one of the two statistical interfaces : https://github.com/nistats/nistats

My provisional plan is to start development on something that might be called nipy2, with a changed interface, fixing some of the painful parts of the current nipy interface.  At that point it would be a good idea to import from those modules, if they are in good shape.

matthew-brett:
Sorry - I should also say that modern versions of the visualization stuff are in https://github.com/nilearn/nilearn

PR #417, opened 2016-09-24
RF: refactor DCT-II into more visible function

matthew-brett:
Make DCT-II basis an API function, use this function for DCT-II drift in
design_matrix.

@jbpoline - would you mind having a look at this one?


matthew-brett:
@bthirion - will the requirement for continuous samples in time be a problem for you?

https://github.com/nipy/nipy/pull/417/files#diff-9967264e8820787d766a0f3ee1e4ae67R42

matthew-brett:
@jbpoline  - any comments?

matthew-brett:
@jbpoline @bthirion - another friendly reminder asking for comments.

matthew-brett:
Gentlemen - I think the current function will be incorrect when the frametimes are not all equally spaced, do you agree?

PR #419, opened 2017-01-22
MRG: compatibility with numpy 1.12.0

matthew-brett:
Numpy 1.12 started raising errors for float indices in a variety of
circumstances, causing lots of test errors - e.g.
https://travis-ci.org/nipy/nipy/jobs/193952659

### DIFF
--- a/nipy/algorithms/clustering/imm.py
+++ b/nipy/algorithms/clustering/imm.py
@@ -4,6 +4,7 @@ with an unspecified number of classes
 """
 from __future__ import print_function
 from __future__ import absolute_import
+from __future__ import division
 
 import math
 
@@ -601,7 +602,7 @@ class MixedIMM(IMM):
         if np.isscalar(kfold):
             aux = np.argsort(np.random.rand(n_samples))
             idx = - np.ones(n_samples).astype(np.int)
-            j = np.ceil(n_samples / kfold)
+            j = int(math.ceil(n_samples / kfold))
             kmax = kfold
             for k in range(kmax):
                 idx[aux[k * j:min(n_samples, j * (k + 1))]] = k
--- a/nipy/algorithms/clustering/imm.py
+++ b/nipy/algorithms/clustering/imm.py
@@ -5,6 +5,8 @@ with an unspecified number of classes
 from __future__ import print_function
 from __future__ import absolute_import
 
+import math
+
 import numpy as np
 
 from scipy.special import gammaln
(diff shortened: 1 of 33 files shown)


nipy/algorithms/clustering/imm.py
```diff
@@ -234,7 +236,7 @@ def cross_validated_update(self, x, z, plike, kfold=10):
         if np.isscalar(kfold):
             aux = np.argsort(np.random.rand(n_samples))
             idx = - np.ones(n_samples).astype(np.int)
-            j = np.ceil(n_samples / kfold)
+            j = int(math.ceil(n_samples / kfold))
```
matthew-brett: Yes, will be an integer in Python 2.   I guess the question for @bthirion is - was this intended to be true (floating point division) even in Python 2?   At the moment this can give different answers on Python 2 and Python 3.

nipy/algorithms/clustering/tests/test_hierarchical_clustering.py
```diff
@@ -157,7 +159,7 @@ def ward_test_more(n=100, k=5, verbose=0):
     # Check that two implementations give the same result
     np.random.seed(0)
     X = randn(n,2)
-    X[:np.ceil(n/3)] += 5
+    X[:int(math.ceil(n / 3))] += 5
```
matthew-brett: @bthirion - how about this one?  Presumably meant to be true division?

bthirion: Indeed, it is n // 3

nipy/algorithms/clustering/imm.py
bthirion: Yes, it is an int. I think that the result is equivalent in Python 2 and 3 ?

effigies: If `kfold` doesn't divide evenly, it will be off by one:

Suppose `n_samples = 5`, `kfold = 2`.

Python 2: `int(math.ceil(5 / 2)) == int(math.ceil(2)) == int(2.0) == 2`
Python 3: `int(math.ceil(5 / 2)) == int(math.ceil(2.5)) = int(3.0) == 3`

nipy/algorithms/clustering/tests/test_hierarchical_clustering.py
matthew-brett: @bthirion - I was thinking you meant this be floating point division, otherwise the `ceil` call doesn't make sense, because the integer division does an implicit `floor` on this positive number.  But you did mean integer division?  In which case we can discard the `ceil`.

nipy/algorithms/clustering/imm.py
bthirion: All right. In that case, we don't really care about the exact value, so both are fine.

nipy/algorithms/clustering/tests/test_hierarchical_clustering.py
bthirion: Same here.

matthew-brett:
Thanks for the review and comments.

PR #421, opened 2017-01-27
TST: relax imm test threshold

### LINKED ISSUE #418: Intermittent test failures in test_imm_loglike_1D_k10
I observed this failure:

```
======================================================================
FAIL: nipy.algorithms.clustering.tests.test_imm.test_imm_loglike_1D_k10
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/mb312/.virtualenvs/test3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/mb312/dev_trees/nipy/nipy/algorithms/clustering/tests/test_imm.py", line 86, in test_imm_loglike_1D_k10
    assert_true(np.absolute(theoretical_ll-empirical_ll)<0.25*dim)
AssertionError: False is not true
    'False is not true' = self._formatMessage('False is not true', "%s is not true" % safe_repr(False))
>>  raise self.failureException('False is not true')
```

Investigating: 

```
(Pdb) np.absolute(theoretical_ll-empirical_ll)
0.25215277379595946
(Pdb) dim
1
```

The test passed when I ran it again without change to the code.  Test sensitive to variation from random number input?

matthew-brett:
Allow for occasional failure due to variation in random number output.

Closes #418.

### DIFF
--- a/nipy/algorithms/clustering/tests/test_imm.py
+++ b/nipy/algorithms/clustering/tests/test_imm.py
@@ -83,7 +83,9 @@ def test_imm_loglike_1D_k10():
     like =  igmm.sample(x, niter=300, kfold=k)
     theoretical_ll = -dim*.5*(1+np.log(2*np.pi))
     empirical_ll = np.log(like).mean()
-    assert_true(np.absolute(theoretical_ll-empirical_ll)<0.25*dim)
+    # Result susceptible to random number output. See:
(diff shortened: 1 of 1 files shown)


matthew-brett:
@bthirion - I think you've already implicitly approved this one, from your comment in #418 - do you agree?

bthirion:
Yes.

Issue #420, opened 2017-01-22
Test failure for mroi.py and Python 3.6


matthew-brett:
Test failure only on Python 3.6.   Maybe a dictionary ordering error?

```
======================================================================
ERROR: Failure: IndexError (list index out of range)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nose/loader.py", line 252, in generate
    for test in g():
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/tests/test_bsa.py", line 84, in test_bsa_methods
    algorithm=name)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/tests/test_bsa.py", line 37, in make_bsa_2d
    algorithm=algorithm, n_iter=100, burnin=10)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/bayesian_structural_analysis.py", line 375, in compute_landmarks
    domain, stats, threshold, smin, method)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/bayesian_structural_analysis.py", line 133, in _compute_individual_regions
    domain, stats_, threshold=threshold, smin=smin)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/hroi.py", line 113, in HROI_as_discrete_domain_blobs
    nroi = hroi_agglomeration(nroi, criterion=criterion, smin=smin)
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/hroi.py", line 71, in hroi_agglomeration
    output_hroi.select_roi(output_hroi.get_id()[value > smin])
  File "/home/travis/build/matthew-brett/nipy/venv/lib/python3.6/site-packages/nipy/labs/spatial_models/hroi.py", line 301, in select_roi
    SubDomains.select_roi(self, id_list)
[...]

PR #422, opened 2017-02-01
BF: fix bug modifying feature dictionary of MROI

matthew-brett:
When iterating through feature dictionary, the hroi module was using the
dictionary keys, which are being modified inside the loop.  Use a static
list of the keys to prevent this happening.

See: https://github.com/nipy/nipy/issues/420

### DIFF
--- a/nipy/labs/spatial_models/mroi.py
+++ b/nipy/labs/spatial_models/mroi.py
@@ -700,14 +700,14 @@ class SubDomains(object):
         # set new features
         # (it's ok to do that after labels and id modification since we are
         # poping out the former features and use the former id indices)
-        for fid in self.features:
+        for fid in list(self.features):
             f = self.remove_feature(fid)
             sf = [f[id] for id in id_list_pos]
             self.set_feature(fid, sf)
         # set new ROI features
         # (it's ok to do that after labels and id modification since we are
         # poping out the former features and use the former id indices)
-        for fid in self.roi_features:
+        for fid in list(self.roi_features):
             if fid != 'id':
                 f = self.remove_roi_feature(fid)
                 sf = np.ravel(f[int(id_list_pos)])
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,6 @@ matrix:
       env:
         - INSTALL_TYPE=requirements
         - DEPENDS=
-  allow_failures:
     - python: 3.6
 
 before_install:
--- a/nipy/labs/spatial_models/tests/test_bsa.py
+++ b/nipy/labs/spatial_models/tests/test_bsa.py
@@ -82,8 +82,8 @@ def test_bsa_methods():
         # run the algo
         AF, BF = make_bsa_2d(betas, theta, sigma, ths, thq, smin,
                              algorithm=name)
-        yield assert_true, test_func(AF, BF)
-    
+        assert_true(test_func(AF, BF))
+
     assert_true(AF.map_label().shape == (np.prod(shape),))
     assert_true(AF.kernel_density().shape == (np.prod(shape),))
     assert_true((AF.roi_prevalence() > ths).all())


bthirion:
The change seems innocuous, so this is good for me.

matthew-brett:
Thanks for the review.

VirgileFritsch:
> the hroi module was using the dictionary keys, which are being modified inside the loop

This should have therefore been a problem with every Python version.
I was thinking that the issue was related to the elements of the dictionnary being selected using their position instead of their key.

nipy/labs/spatial_models/mroi.py
```diff
@@ -700,14 +700,14 @@ def select_roi(self, id_list):
         # set new features
         # (it's ok to do that after labels and id modification since we are
         # poping out the former features and use the former id indices)
-        for fid in self.features:
+        for fid in list(self.features):
             f = self.remove_feature(fid)
             sf = [f[id] for id in id_list_pos]
```
VirgileFritsch: `id_list_pos` is a list of "positions" of the dictionnary elements that should be looped over. This code should break if the order of the dictionnary elements cannot be relied upon.

nipy/labs/spatial_models/mroi.py
```diff
@@ -700,14 +700,14 @@ def select_roi(self, id_list):
...
             self.set_feature(fid, sf)
         # set new ROI features
         # (it's ok to do that after labels and id modification since we are
         # poping out the former features and use the former id indices)
-        for fid in self.roi_features:
+        for fid in list(self.roi_features):
             if fid != 'id':
                 f = self.remove_roi_feature(fid)
                 sf = np.ravel(f[int(id_list_pos)])
```
VirgileFritsch: I am being doubtful about `id_list_pos` here again.

matthew-brett:
Virgile - did you read http://bugs.python.org/issue29420? - it records the difference in iteration behavior between the Python versions.   The basic message is - modifying the dictionary while iterating generates undefined behavior - and that undefined behavior changed.

nipy/labs/spatial_models/mroi.py
matthew-brett: Virgile - it looks to me as if the output from `f = self.remove_feature(fid)` is a list.  So maybe the `fid` is badly named, it should be called `feature_name` where the feature value is a list of arrays.  Is that right?

VirgileFritsch:
I just did. Still a bit confused by the "difference-in-interation" issue and the "no-order-guarantied" one but it does not matter as far as the fix is ok.

Maybe you want to stay aware that somewhere in the mroi code, we access to the values of a dictionnary by their positions and not their key. It seems to me that this is not a good thing... but I might be wrong :)

Thank you @matthew-brett and @bthirion for taking care of the issue.

matthew-brett:
Can you find a place in the code where we are accessing the items in the dictionary by their position?   I did have a quick look, but found no obvious place.

matthew-brett:
The failure here was caused entirely by the difference in iteration behavior - hence the fix, which is to stop the loop modifying what gets iterated over.

VirgileFritsch:
> Can you find a place in the code where we are accessing the items in the dictionary by their position?

I did in the diff ;)

matthew-brett:
Did you get my comment on your comment?   I think you are wrong about `id_list_pos` referring to positions of the dictionary keys, I think it refers to positions within the lists that are the `feature` values.

nipy/labs/spatial_models/mroi.py
VirgileFritsch: Everything is right. Ok for renaming `fid` into `feature_name`, although I think it is more interesting to rename `f` into ... `current_feature`?

VirgileFritsch:
Right, I was mistaken.
I remember a part of the code *did* rely on the dictionnary order, but maybe I or someone fixed it a long time ago. Everything seems fine then.

PR #423, opened 2017-02-02
RF: rename variables in ROI selection loops

matthew-brett:
Rename variables for clarity, as proposed in
https://github.com/nipy/nipy/pull/422

### DIFF
--- a/nipy/labs/spatial_models/mroi.py
+++ b/nipy/labs/spatial_models/mroi.py
@@ -69,6 +69,9 @@ class SubDomains(object):
           should access ROI through their id to avoid hazardous manipulations.
 
         """
+        self._init(domain, label, id)
+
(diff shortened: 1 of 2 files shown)


matthew-brett:
By the way, I think https://github.com/nipy/nipy/blob/master/nipy/labs/spatial_models/mroi.py#L689 must be wrong - the `self` variable will get rebound, and so have no effect on the `self` in the method signature.  You'd need to set the properties / attributes of `self` directly.

matthew-brett:
I added a fix and test for the select_roi bug.

matthew-brett:
@VirgileFritsch - any comment here?

VirgileFritsch:
That's fine. Thank you Matthew :)

matthew-brett:
Thanks - in it goes.

PR #425, opened 2017-02-06
BF: refix the select_roi indexing

matthew-brett:
My original indexing fix only worked for selection of one ROI.

Add new tests from the example from which I found the error.

The example was::

    python examples/labs/need_data/demo_blob_from_image.py


matthew-brett:
@VirgileFritsch - sorry - I found a problem with one of my fixes.  Can you have a look at the fix and the test?  For example, I see that the ROI code is casting to int when collecting features, which has an odd effect for the activation feature.

matthew-brett:
Sorry - this really is the last one that needs reviewing before merge / release.

matthew-brett:
@VirgileFritsch - can I direct your attention to https://github.com/nipy/nipy/pull/425/files#diff-5915411cade2c175a3ad593e220c4c24R716 in particular?   Does the tested output look right in the new test?

matthew-brett:
Well - please do comment if you have the time.  I'll merge in any case, but if it's wrong, please let me know.

PR #426, opened 2017-02-06
BF: fix numpy 1.12 indexing error in gmm plot

matthew-brett:
gmm plot attempting reshape with float element.


matthew-brett:
@bthirion - I think this one must be right.

bthirion:
The code seems correct to me.

matthew-brett:
Thanks for the review.

PR #427, opened 2017-03-05
MRG: test against pre-release numpy / scipy

matthew-brett:
Use pre-release numpy / scipy wheels for testing.


matthew-brett:
Any comments here?

Issue #448, opened 2018-12-19
Question regarding temporal filtering


takhs91:
Hi,
I was  reading the way temporal filtering is implemented using the cosine model and except I was a bit confused from the fact that is written as a low-pass filter while it is a high-pass filter as also stated in https://github.com/nipy/nipy/issues/295 (An error that is actually propagated in nistats too https://github.com/nistats/nistats/blob/master/nistats/design_matrix.py#L87) I also think I noticed something else.

The order of the DCT that is computed at https://github.com/nipy/nipy/blob/master/nipy/modalities/fmri/design_matrix.py#L87
is the closest one  (equal to  or a bit lower in frequency ) to the requested cutoff point.

For example if you have 120 volumes and a tr of 2s and you want your cut period to be 120s then highest order should be 

`2*120*(1/120)*2 = 4`

this is the exact dct order that will cut the requested period

Although later https://github.com/nipy/nipy/blob/master/nipy/modalities/fmri/design_matrix.py#L91
only the 1... order-1 drifts are added (plus the constant at last)missing the dct of order equal to the order found above.

Should't it be included too? If that's the case I am willing to proceed with a fix for here and for nistats.

bthirion:
Thx for the comments. 
Note that here, ou plan is to stray away from SPm conventions and provide the frequency in Herz. 
Note thate the cosine drift provides a basis for a low-pass filter of the data; the fact that it is treated as a nuisance parameters makes the whole thing a high-pass filtering. We should probably coroect the wording (PR welcome ---in nistats: the orriginal nipy code is dead). 
Regarding whether we should consider the last regressor or not, it depends whether you consider the noud as incluisve or exclusive ;-) We took the conservative option, but I have no strong opinion on that. 
Best, 
B

takhs91:
Thanks for replying,
the documentation `Create a cosine drift matrix with periods greater or equals to period_cut` gave me the idea its should be inclusive.

Except from that though in the case when the calculated order is exact to the one matching the cutt-of, skipping it is the conservative option.

On the other hand at the cases where the order is not exact (where using the floor func makes the difference) using that order would be already conservative. Skipping it is not necessary and makes it too conservative.

For example, using the same data as in the example above but using a cutoff of 100s

order = floor(2*120*(1/100)*2) = floor(4,8)= 4

in that case using 4 would be the conservative choice but using 3 (this is the default now) is too conservative.

If skipping is kept maybe using round instead of floor would be more even.

[...]

PR #449, opened 2018-12-20
Add highest dct order

takhs91:
Spawn from #448


takhs91:
At the tests where you compare against spm drifts I had to exclude the highest order since it seems that spm was also didn't add the last possible. I don't use spm though so I am not sure if that is configurable or not. If it is a new file with equal number of drifts could be produced

takhs91:
Any thoughts on that?

takhs91:
the relevant PR was merged at nistats https://github.com/nistats/nistats/pull/301

Any updates for here?

matthew-brett:
Sorry - I'm afraid I have been crazy busy these last weeks.  I will try very hard to get to this at the weekend.  Thanks for the reminder.

takhs91:
Hi! Any updates on that?

PR #452, opened 2019-05-21
Remove deprecated references

Czorio:
As of Scipy 1.0.0 the factorial method no longer resides in scipy.misc, but it is found in scipy.special instead. I have fixed the imports in the files that I could find that try to import factorial from scipy.misc.

### DIFF
--- a/nipy/algorithms/statistics/rft.py
+++ b/nipy/algorithms/statistics/rft.py
@@ -20,8 +20,7 @@ import numpy as np
 from numpy.linalg import pinv
 
 from scipy import stats
-from scipy.misc import factorial
-from scipy.special import gamma, gammaln, beta, hermitenorm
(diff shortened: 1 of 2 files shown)


jcohenadad:
oopsi! i missed that PR. Well, I created another one here: https://github.com/nipy/nipy/pull/454
One notable difference between this PR and my PR is that the call to `scipy.misc.comb` also needs to be updated.

Czorio:
That's fine, I only ran into this issue because it broke some functionality in the spinal cord toolkit and had to dig around in the included packages, so I figured I might as well toss my changes to you guys, too.

jcohenadad:
@Czorio while we wait for the nipy team to review the PR, we have redirected the nipy dependency in SCT to a temporary fork: https://github.com/neuropoly/spinalcordtoolbox/pull/2269

PR #469, opened 2021-02-18
shebang to python3 not python

yarikoptic:
As on some platforms python might still be pointing to good old
python2 and I would expect python3 present - use it in shebang
to signify need for python3

### DIFF
--- a/doc/devel/tools/tricked_out_emacs.rst
+++ b/doc/devel/tools/tricked_out_emacs.rst
@@ -83,7 +83,7 @@ Install pylint_.  Debian packages pylint_ as ``pylint``. Put the
 `flymake .emacs snippet`_ in your ``.emacs`` file.  You will see, in the
 emacs_python_mode_ page, that you will need to save this::
 
-    #!/usr/bin/env python
+    #!/usr/bin/env python3
     
     import re
     import sys
--- a/doc/sphinxext/autosummary_generate.py
+++ b/doc/sphinxext/autosummary_generate.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 r"""
--- a/examples/affine_registration.py
+++ b/examples/affine_registration.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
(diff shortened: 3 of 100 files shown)


effigies:
Last I knew Nipy still claimed Py2 support. Should the next release do away with that?

matthew-brett:
Yes, that seems a little extreme.   I personally alias python to python3 - wouldn't we expect most of our users to do something like that, if that's what they wanted?

yarikoptic:
Are you suggesting to include into installation instructions necessity to alias python to python 3? (Bash alias isn't enough BTW for shebangs afaik, so need to be a symlink... System wide?)

What are you gaining by keeping ability to run with EOLed python 2?

yarikoptic:
> Last I knew Nipy still claimed Py2 support. Should the next release do away with that?

It could as well support python 1 . Here is the question with what python to run scripts on a system after installation. Debian got rid of python (2). We can keep the patch for Debian package but then likely people wouldn't be able to run them if installed otherwise

effigies:
So Debian now has `python3` but not `python`? Part of the point of writing scripts that supported both Python 2 and 3 was so that you didn't have to care what the local Python was. Scripts that depended on a particular version could put `python2` or `python3`.

I'm for dropping Python 2, but I'd rather do it as a deliberate act, e.g., set it in the classifiers and set `python_requires` so that Python 2 installations at least have a chance of not installing something that is guaranteed to break.

matthew-brett:
I think #473 is a duplicate.

Is this a Debian requirement / default?   I don't care enough about it to resist, if it is.

mr-c:
For the next release of Debian (which is being finalized now), only `python3` is supported. `/usr/bin/python`  might exist, but it is a separate package that might go away.

matthew-brett:
It is right to do this in the `scripts` directory?  Won't these be rewritten by installation?

mr-c:
> I'm for dropping Python 2, but I'd rather do it as a deliberate act, e.g., set it in the classifiers and set `python_requires` so that Python 2 installations at least have a chance of not installing something that is guaranteed to break.

I recommend this checklist https://adamj.eu/tech/2019/03/13/dropping-python-2-support/

Followed by `git ls-files | grep py$ | xargs pyupgrade --py36-plus`

matthew-brett:
Thanks - that's useful.   I think we should keep Python 2 support for this next release though - as the last release supporting Python 2.

effigies:
If we're not breaking py2 immediately, then I'd say the cleanest way to achieve this task would be to move scripts to entrypoints, and the install process should ensure they have the right shebang. That should work, since I haven't heard complaints about nibabel scripts, and that's in neurodebian if not Debian.

matthew-brett:
@effigies - good idea - working on that over at #476

PR #473, opened 2021-02-20
fix python invocation to be specifically python3

mr-c:


matthew-brett:
We're discussing this over at #469 .  Could we move over there?

PR #477, opened 2021-02-22
MRG: remove deprecated use of np.object

matthew-brett:
Apparently "object" is identical in behavior.


effigies (approved):
Seems clearly safe.

matthew-brett:
Thanks Chris.

Issue #468, opened 2021-02-18
test_interpolator fails with quite a swing: -3.4 vs 3


yarikoptic:
```
smaug:/tmp/buildd/nipy-0.4.3~rc1# python3 -m nose .pybuild/cpython3_3.9_nipy/build/nipy/algorithms/tests/test_interpolator.py:test_interpolator
nipy.algorithms.tests.test_interpolator.test_interpolator ... FAIL

======================================================================
FAIL: nipy.algorithms.tests.test_interpolator.test_interpolator
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/build/nipy-0.4.3~rc1/.pybuild/cpython3_3.9_nipy/build/nipy/algorithms/tests/test_interpolator.py", line 23, in test_interpolator
    assert_almost_equal(interp.evaluate([0, 0, 4]), arr[0, 0, -1])
  File "/usr/lib/python3/dist-packages/numpy/testing/_private/utils.py", line 579, in assert_almost_equal
    return assert_array_almost_equal(actual, desired, decimal, err_msg)
  File "/usr/lib/python3/dist-packages/numpy/testing/_private/utils.py", line 1042, in assert_array_almost_equal
    assert_array_compare(compare, x, y, err_msg=err_msg, verbose=verbose,
  File "/usr/lib/python3/dist-packages/numpy/testing/_private/utils.py", line 840, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 7 decimals

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 6.4
Max relative difference: 2.13333333
 x: array(-3.4)
 y: array(3)
>>  raise AssertionError('\nArrays are not almost equal to 7 decimals\n\nMismatched elements: 1 / 1 (100%)\nMax absolute difference: 6.4\nMax relative difference: 2.13333333\n x: array(-3.4)\n y: array(3)')
```

yarikoptic:
relates to upgrade of scipy from 1.5.4-1+b1 to 1.6.0-2

yarikoptic:
[...]

PR #478, opened 2021-02-23
MRG: fix new interpolation errors with Scipy 1.6

matthew-brett:
Scipy >= 1.6.0 not longer guarantees that you get the same result when
interpolating directly (with map_coordinates, prefilter=True - the default) and
when using pre-calculated filters (map_coordinates, prefilter=False).

Therefore, we have to pad the calculated filter to get the same result.

See:

* https://github.com/nipy/nipy/issues/468
* https://github.com/scipy/scipy/issues/13600

### DIFF
--- a/nipy/algorithms/interpolation.py
+++ b/nipy/algorithms/interpolation.py
@@ -15,6 +15,16 @@ from ..fixes.scipy.ndimage import map_coordinates
 from ..utils import seq_prod
 
 
+# Earlier versions of Scipy don't have mode for spline_filter
+SPLINE_FILTER_HAS_MODE = True
+# Ensure bare call does not cause error.
+ndimage.spline_filter(np.zeros((2,2)))
+try:  # Does adding mode cause error?
+    ndimage.spline_filter(np.zeros((2,2)), mode='constant')
+except TypeError:
+    SPLINE_FILTER_HAS_MODE = False
+
+
 class ImageInterpolator(object):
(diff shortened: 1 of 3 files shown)


matthew-brett:
Any chance of a quick review here?  With #479, I think the tests will be (finally) passing.

matthew-brett:
Yes, looks this one will have the tests passing again.

yarikoptic:
wooohoo -- well done @matthew-brett !

matthew-brett:
Last call for a review here?   I plan to merge soon to get the tests passing.

effigies:
I don't see anything wrong, but I also don't work enough with scipy interpolation to have a good intuition of where a mistake might be. So in the absence of a more knowledgeable review, I'll say the Python looks reasonable and the tests have good coverage, so LGTM.

matthew-brett:
Thanks Chris - let's give this one a shot to get the tests passing.  I'll add some more tests to the interpolation in another PR.

matthew-brett:
Writing some more tests got me worried : https://github.com/scipy/scipy/issues/13600#issuecomment-786322347

yarikoptic:
Tests are the pain of our lives! Let's stop testing to get back to the happy blissful ignorance ;-)

matthew-brett:
Yes - testing - horrible.   But, now I've spend some time down there : https://github.com/nipy/nipy/pull/481

PR #479, opened 2021-02-23
BF: change in np.prod upsets Formulae

matthew-brett:
As of Numpy 1.20, np.prod calls the __array__ method of input objects,
and this messes up Formula products.  Adapt and use basic sequence
product instead.


matthew-brett:
Also plea for a quick review, to get the tests passing (I think tests pass after we've merged this and #478).

effigies (approved):
Makes sense.

matthew-brett:
Thanks much for getting to this so fast.

PR #480, opened 2021-02-24
MRG: example etc shebangs to Python3

matthew-brett:
Debian may only install Python 3.

Based on, supercedes #469.


yarikoptic (approved):

yarikoptic:
approve, didn't look into CI fails

matthew-brett:
I've fixed some relevant CI fails, the others are now from the bug fixed in #478 - so I think we're nearly there now.

PR #481, opened 2021-03-02
MRG: Extend testing, fix / simplify interpolation

matthew-brett:
Extend testing of prefiltered interpolation.

This revealed various bugs:

* Prefiltering=False of linear interpolation was failing, because the
  prefilter=False option didn't make sense - because we were (correctly)
  not prefiltering.
* Use more standard TemporaryFile for backing store to avoid errors from
  too many open files.
* Avoid padding before prefiltering for Scipy < 1.6, because this gives
  different results from non-prefiltered case.

Also see #478

### DIFF
--- a/nipy/algorithms/interpolation.py
+++ b/nipy/algorithms/interpolation.py
@@ -4,11 +4,13 @@
 """
 from __future__ import absolute_import
 
-import os
 import tempfile
+
+from distutils.version import LooseVersion as LV
 
 import numpy as np
 
+import scipy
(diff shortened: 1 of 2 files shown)


matthew-brett:
I think this is the right fix, with extra tests, for the issue in #478 - as well as a couple of extra issues I found during testing.

matthew-brett:
Last call for a review here?

effigies (approved):

matthew-brett:
Thanks Chris.

Issue #464, opened 2020-11-03
New release?


mr-c:
Debian would appreciate a proper new release of nipy; it is in danger of being removed from Debian completely (it has already been removed from our testing distribution) 

https://bugs.debian.org/959138

Thanks!

effigies:
@mr-c Does Debian pull from the GitHub archive or from PyPI? And do you know if current master is in a state where it can be packaged for Debian?

mr-c:
Currently we watch for new tags on GitHub, but we can also watch PyPI

https://salsa.debian.org/med-team/nipy/-/blob/master/debian/watch

mr-c:
> And do you know if current master is in a state where it can be packaged for Debian?

If someone cuts and tags a pre-release then I'll upload an updated package to Debian's "experimental" distribution.

effigies:
@mr-c I just tagged 0.4.3rc1.

mr-c:
Thanks @effigies ! I just opened a new issue with the results on my initial attempt at updating the Debian package for 0.4.3rc1: https://github.com/nipy/nipy/issues/466

matthew-brett:
I've just released 0.5.0 - @mr-c - can you check it works for you?

mr-c:
Seems so, though I get an error with building the docs; perhaps because you don't allow the use of a recent sphinx version? Debian is on Sphinx 3.4.3 ; `doc-requirements.txt` lists `sphinx>=1.0,<3.0`

matthew-brett:
See https://github.com/nipy/nipy/pull/485

@mvdoc - did I miss something in allowing updates to Sphinx and Numpydoc?  Sorry, I didn't go back to check what you did before.

mr-c:
Thanks @matthew-brett ; https://github.com/nipy/nipy/pull/485/files#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R19 is what I also figured out locally

mr-c:
Hmm.. now we can't run `nipy.tests.test_scripts.test_{nipy_3_4,nipy_4d_realign,nipy_tsdiffana,nipy_diagnose}` as `run_command` makes assumptions about the file system layout and overwrites the `PYTHONPATH` we were setting for the running of the tests..

```
======================================================================
ERROR: nipy.tests.test_scripts.test_nipy_4d_realign
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #485, opened 2021-03-30
Relax requirements for documentation tools

matthew-brett:
Docs appear to build OK with latest Sphinx and Numpydoc

See #464

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -1,8 +1,8 @@
 # Requirements for building docs
 # Check these dependencies against doc/conf.py
 -r dev-requirements.txt
-sphinx>=1.0,<3.0
-numpydoc==0.8.0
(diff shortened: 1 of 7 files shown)


doc-requirements.txt
```diff
@@ -1,8 +1,8 @@
 # Requirements for building docs
 # Check these dependencies against doc/conf.py
 -r dev-requirements.txt
-sphinx>=1.0,<3.0
-numpydoc==0.8.0
+sphinx>=1.0
+numpydoc>=0.8.0
```
jarrodmillman, line 5:
```suggestion
sphinx>=7.0
numpydoc>=1.6.0
```

I don't think there is any reason that you would need to support older versions for doc building.

jarrodmillman (approved):

PR #487, opened 2021-06-20
[MAINT] More Depreciation Changes

alexrockhill:
Sphinx was giving me errors including `nipy` as a dependency because of the depreciation warnings so I took a bit and changed enough to get rid of the warnings. I know there is a lot more to do so maybe this is too little changes but I thought since I did it, I would post a PR at least.

Also all the tests pass so I think it works okay.


alexrockhill:
Hmm so this let me build my documentation but seems like I changed quite a few things too much so probably just better to close.

PR #488, opened 2021-06-20
[MAINT] Fixes some warnings

alexrockhill:
This is a scaled back version of depreciation warnings that caused issues with Sphinx.


effigies:
Looks like this got done in other PRs.

PR #498, opened 2023-02-02
Account for nibabel 5.0.0 removal of .py3k shim - use numpy.compat.py3k

yarikoptic:
Includes some other fixes found in the patch prepared by @emollier (thanks!) in Debian:
https://salsa.debian.org/med-team/nipy/-/blob/master/debian/patches/nibabel5.0.0.patch

### DIFF
--- a/nipy/core/image/image_spaces.py
+++ b/nipy/core/image/image_spaces.py
@@ -60,7 +60,7 @@ array([[ 2.,  0.,  0.,  0.],
 It also works with nibabel images, which can only have xyz_affines:
 
 >>> import nibabel as nib
->>> nimg = nib.Nifti1Image(data, affine)
+>>> nimg = nib.Nifti1Image(data, affine, dtype=data.dtype)
 >>> xyz_affine(nimg)
 array([[ 2.,  0.,  0.,  0.],
        [ 0.,  3.,  0.,  0.],
@@ -132,7 +132,7 @@ def xyz_affine(img, name2xyz=None):
     Nibabel images always have xyz affines
 
     >>> import nibabel as nib
-    >>> nimg = nib.Nifti1Image(arr, np.diag([2,3,4,1]))
+    >>> nimg = nib.Nifti1Image(arr, np.diag([2,3,4,1]), dtype=arr.dtype)
     >>> xyz_affine(nimg)
     array([[ 2.,  0.,  0.,  0.],
            [ 0.,  3.,  0.,  0.],
@@ -195,7 +195,7 @@ def is_xyz_affable(img, name2xyz=None):
     Nibabel images always have xyz affines
 
     >>> import nibabel as nib
-    >>> nimg = nib.Nifti1Image(arr, np.diag([2,3,4,1]))
+    >>> nimg = nib.Nifti1Image(arr, np.diag([2,3,4,1]), dtype=arr.dtype)
     >>> is_xyz_affable(nimg)
     True
     """
--- a/nipy/labs/spatial_models/parcel_io.py
+++ b/nipy/labs/spatial_models/parcel_io.py
@@ -54,7 +54,8 @@ def mask_parcellation(mask_images, nb_parcel, threshold=0, output_image=None):
         # mask_images should be a list
         mask_data = intersect_masks(mask_images, threshold=0) > 0
         mask = Nifti1Image(mask_data.astype('u8'),
-                           get_affine(load(mask_images[0])))
+                           get_affine(load(mask_images[0])),
+                           dtype=mask_data.astype('u8').dtype)
 
     domain = grid_domain_from_image(mask)
     cent, labels, J = kmeans(domain.coord, nb_parcel)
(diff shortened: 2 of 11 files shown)


nipy/core/image/image_spaces.py
```diff
@@ -60,7 +60,7 @@
 It also works with nibabel images, which can only have xyz_affines:
 
 >>> import nibabel as nib
->>> nimg = nib.Nifti1Image(data, affine)
+>>> nimg = nib.Nifti1Image(data, affine, dtype=data.dtype)
```
effigies, line 63: This will set a hard minimum of `nibabel >= 4.0`. The alternative is to set a non-int64 dtype on data (L21).

nipy/labs/spatial_models/parcel_io.py
```diff
@@ -100,7 +101,8 @@ def parcel_input(mask_images, learning_images, ths=.5, fdim=None):
         # mask_images should be a list
         grp_mask = intersect_masks(mask_images, threshold=ths) > 0
         mask = Nifti1Image(grp_mask.astype('u8'),
-                           get_affine(load(mask_images[0])))
+                           get_affine(load(mask_images[0])),
+                           dtype=grp_mask.astype('u8').dtype)
```
effigies, line 105: I suspect this was meant to be `uint8` (`u1`), and instead it's `uint64` (`u8`). Why would you insist on creating a mask of 64-bit ints?

nipy/labs/spatial_models/tests/test_discrete_domain.py
```diff
@@ -130,7 +130,7 @@ def test_image_domain():
     toto = np.ones(shape[:3])
     affine = np.random.randn(4, 4)
     affine[3:, 0:3] = 0
-    nim = Nifti1Image(toto, affine)
+    nim = Nifti1Image(toto, affine, dtype=toto.dtype)
```
effigies, line 133: Was this causing a problem? `toto` should be float64.

effigies:
The above comments apply in several places.

emollier:
Thanks for your review, upon verification, that code section was
not causing any particular problem.  This is probably the case
for the next section involving the variable `toto` as well.

I initially went through the patch after a couple of iterations,
so might have overdone things near the end.

emollier:
Chris Markiewicz, on 2023-02-02:
> +>>> nimg = nib.Nifti1Image(data, affine, dtype=data.dtype)
> 
> This will set a hard minimum of `nibabel >= 4.0`.

Acknowledged, the patch was initially written to target the
upcoming Debian bookworm, for which we can safely assume that
the nibabel version will be >= 5.0.  But I agree in it's current
form it may not be well suited for use as-is upstream without
some version check to support legacy versions on older operating
systems.

> The alternative is to set a non-int64 dtype on data (L21).

This sounds interesting and much simpler than the current
implementation.  IIUC that would be implemented with the astype
method.  Which dtype would you suggest?

emollier:
I kind of thought the same, but didn't feel comfortable enough
with the codebase to make such change myself if, in the end, it
were intended.  IIUC, the mask reduction would resolve more
appropriately the nibabel port than the current workaround?

effigies:
> > The alternative is to set a non-int64 dtype on data (L21).
> 
> This sounds interesting and much simpler than the current implementation. IIUC that would be implemented with the astype method. Which dtype would you suggest?

Since it's just example data and doesn't specifically need integer properties, `float32` seems reasonable.

effigies:
> IIUC, the mask reduction would resolve more appropriately the nibabel port than the current workaround?

To be fair, I've never touched nipy (I work with nibabel quite a lot), but I think so.

emollier:
Hi Chris, Hi Yaroslav,

I modified the [Debian nipy patch for nibabel 5.0.0] to account
for your recommendtations.  Yaroslav, per chance, do you have
the spare cycles to update the merge request?  Otherwise I will
have to find a way to hook the the existing branch or make a new
one.

[Debian nipy patch for nibabel 5.0.0]: https://salsa.debian.org/med-team/nipy/-/blob/12a4fbea8c99c1e5dc07ee81bc3da1a450617050/debian/patches/nibabel5.0.0.patch

Thank you both for your time!

Have a nice day,  :)
-- 
Étienne Mollier ***@***.***>
Fingerprint:  8f91 b227 c7d6 f2b1 948c  8236 793c f67e 8f0d 11da
Sent from /dev/pts/2, please excuse my verbosity.
On air: Xandria - Euphoria

musicinmybrain:
> https://salsa.debian.org/med-team/nipy/-/blob/12a4fbea8c99c1e5dc07ee81bc3da1a450617050/debian/patches/nibabel5.0.0.patch

I have (rebased and) tested the linked patch in Fedora 38/Rawhide, which also has nibabel 5.0.0. It seems to work for me.

(I do have to add a dependency on `pytest` since `nibabel.testing` imports it. Maybe that should be done in the `python-nibabel` package.)

effigies:
> (I do have to add a dependency on `pytest` since `nibabel.testing` imports it. Maybe that should be done in the `python-nibabel` package.)

Probably. We should not be mandatorily importing pytest, though. I'll need to verify that for the next release.

PR #501, opened 2023-02-06
Remove a couple of unnecessary bundled dependencies

musicinmybrain:
- `configobj` is not used anywhere
- `argparse` is present in the standard library from Python 2.7 onwards


musicinmybrain:
If you are interested in converting `six` and `transforms3d` from bundled dependencies to normal dependencies, I’m happy to submit a follow-up PR for that.

matthew-brett:
Six - yes.   Transforms3d - I have some memory that the bundled copy differs from the current one - but I can go back and check that.

PR #503, opened 2023-02-07
Explicitly import sphinx.ext.doctest

musicinmybrain:
In later versions of Sphinx, we cannot assume that we can use “sphinx.ext.doctest.doctest” after “import sphinx”.

I know that the Sphinx version is upper-bounded, but this does no harm and helps with forward-compatibility.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,6 +16,7 @@ import sys, os
 from importlib import import_module
 
 import sphinx
+import sphinx.ext.doctest
 
 # Doc generation depends on being able to import project
 project = 'nipy'


effigies:
I think this was a bad use of `sphinx.ext.doctest`. We should just import `doctest` and use that instead of `sphinx.ext.doctest.doctest`.

effigies:
I don't know. This isn't my project, I just get notifications. I was hoping to be helpful in passing, but I have neither authority nor much time to play and help get things back to a working state.

musicinmybrain:
Hmm. I think I’m going to drop this for now since I don’t intend to put in the work to get the documentation actually building again. I’ll avoid removing any relevant branches in my fork in case somebody decides to pick up the effort later and finds them useful.

PR #512, opened 2023-09-21
Account for nibabel 5.0.0 removal of .py3k shim - use numpy.compat.py3k

musicinmybrain:
Rebased from:

https://salsa.debian.org/med-team/nipy/-/blob/12a4fbea8c99c1e5dc07ee81bc3da1a450617050/debian/patches/nibabel5.0.0.patch

This is a version of https://github.com/nipy/nipy/pull/498 that applies cleanly and seems to work as expected in Fedora Linux. Please see the original PR for discussion.

### DIFF
--- a/nipy/algorithms/statistics/formula/tests/test_formula.py
+++ b/nipy/algorithms/statistics/formula/tests/test_formula.py
@@ -18,7 +18,7 @@ from .. import formulae as F
 from ..formulae import terms, Term
 from nipy.utils import VisibleDeprecationWarning
 
-from nibabel.py3k import asbytes
+from numpy.compat.py3k import asbytes
 from nibabel.testing import assert_dt_equal
 
 from nose.tools import (assert_true, assert_equal, assert_false,
--- a/nipy/algorithms/statistics/formula/tests/test_formula.py
+++ b/nipy/algorithms/statistics/formula/tests/test_formula.py
@@ -18,7 +18,6 @@ from .. import formulae as F
 from ..formulae import terms, Term
 from nipy.utils import VisibleDeprecationWarning
 
-from numpy.compat.py3k import asbytes
 from nibabel.testing import assert_dt_equal
 
 from nose.tools import (assert_true, assert_equal, assert_false,
@@ -364,7 +363,7 @@ def test_formula_inputs():
         f = F.Factor('myname', levels)
         assert_equal(f.levels, level_names)
     # Sending in byte objects
-    levels = [asbytes(L) for L in level_names]
+    levels = [L.encode() for L in level_names]
     f = F.Factor('myname', levels)
     assert_equal(f.levels, level_names)
 
--- a/nipy/core/image/image_spaces.py
+++ b/nipy/core/image/image_spaces.py
@@ -60,7 +60,7 @@ array([[ 2.,  0.,  0.,  0.],
 It also works with nibabel images, which can only have xyz_affines:
 
 >>> import nibabel as nib
->>> nimg = nib.Nifti1Image(data, affine)
+>>> nimg = nib.Nifti1Image(data.astype('float'), affine)
 >>> xyz_affine(nimg)
 array([[ 2.,  0.,  0.,  0.],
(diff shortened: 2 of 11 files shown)


nipy/algorithms/statistics/formula/tests/test_formula.py
```diff
@@ -18,7 +18,7 @@
 from ..formulae import terms, Term
 from nipy.utils import VisibleDeprecationWarning
 
-from nibabel.py3k import asbytes
+from numpy.compat.py3k import asbytes
```
effigies, line 21: IIRC numpy is deprecating this as well. We should just change `asbytes(s)` to `s.encode()`.

nipy/io/tests/test_image_io.py
```diff
@@ -45,7 +45,7 @@ def test_badfile():
     # nibabel prior 2.1.0 was throwing a ImageFileError for the not-recognized
     # file type.  >=2.1.0 give a FileNotFoundError.
     try:
-        from nibabel.py3k import FileNotFoundError
+        from numpy.compat.py3k import FileNotFoundError
     except ImportError:
         FileNotFoundError = IOError
```
effigies, line 50: Should just remove this. `FileNotFoundError` is now a builtin.

tools/run_log_examples.py
```diff
@@ -25,7 +25,7 @@
 from subprocess import Popen, PIPE
 import re
 
-from nibabel.py3k import asstr
+from numpy.compat.py3k import asstr
```
effigies, line 28: Likewise, here we should just convert `asstr(s)` to `s.decode()`

nipy/algorithms/statistics/formula/tests/test_formula.py
musicinmybrain: Seems logical. Adjusted.

nipy/io/tests/test_image_io.py
musicinmybrain: I assume this package is no longer attempting to support Python 2.7?

effigies: Given that there's hardly maintenance effort to keep it running, extending what little there is to 2.7 seems silly. I don't call the shots here, so I might get overruled, but it seems safe to drop Python 2.

effigies (approved):

nipy/io/tests/test_image_io.py
matthew-brett: Yup - no need to maintain Python 2.7 at this stage.

nipy/core/image/image_spaces.py
```diff
@@ -60,7 +60,7 @@
 It also works with nibabel images, which can only have xyz_affines:
 
 >>> import nibabel as nib
->>> nimg = nib.Nifti1Image(data, affine)
+>>> nimg = nib.Nifti1Image(data.astype('float'), affine)
```
matthew-brett, line 63: Sorry - remind me why we need to cast as float now?

matthew-brett (commented):
One point of clarification.

nipy/core/image/image_spaces.py
effigies: It's a quick hack. The problem is that `np.arange()` (used to generate synthetic data) produces `int64`s and now we error on int64 if we don't get an explicit `header` or `dtype` parameter.

* https://github.com/nipy/nibabel/issues/1089

matthew-brett: Worth a comment to that effect?

effigies: I think a comment in a docstring would be more distracting than illuminating. Better solution, IMO, would be to update the `data = np.arange(...)` to be `data = np.arange(..., dtype='float32')`.

matthew-brett: Yes, that would also be fine.

musicinmybrain: I pushed a follow-up commit that I think should be satisfactory.

PR #516, opened 2023-09-21
MNT: Update build requirements to use numpy 1.25 and cython<3

### LINKED ISSUE #508: Fails to build with Cython 3.0.2
It looks like `nipy` (current `master`, https://github.com/nipy/nipy/commit/7eede02471567487e454016c1e7cf637d3afac9e) needs some work to be compatible with [Cython 3](https://cython.readthedocs.io/en/latest/src/changes.html#major-themes-in-3-0-0).

```
python3.11 -m venv _e
. _e/bin/activate
pip install -r requirements.txt
pip install Cython
python setup.py build_ext
```

```
Error compiling Cython file:
------------------------------------------------------------
...
    size = X.shape[0]
    stride = X.strides[0]
    itemsize = X.descr.elsize
    type = X.descr.type_num 
    
    relstride = stride/itemsize
                      ^
------------------------------------------------------------

nipy/labs/bindings/wrapper.pyx:91:22: Cannot assign type 'double' to 'int'
error: 1 errors while compiling 'nipy/labs/bindings/wrapper.pyx' with Cython
```

```
pip install 'Cython<3'
python setup.py build_ext
```

(Now the build is successful.)

effigies:
Fixes #508 and simplifies the build logic a bit, thanks to backwards-compatible numpy builds (see https://numpy.org/doc/stable/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,10 @@
 [build-system]
 requires = [
   "setuptools",
-  "cython",
+  "cython<3",
   # Newer than NEP29-minimum: compile against oldest numpy available
-  "numpy==1.24; python_version >= '3.11'",
-  "numpy==1.22; python_version >= '3.10' and python_version < '3.11'",
-  # NEP29-minimum as of Jan 31, 2023
-  "numpy==1.21; python_version >= '3.7' and python_version < '3.10'",
+  "numpy>=1.25; python_version > '3.8'",
+  # NEP29-minimum as of Sep 21, 2023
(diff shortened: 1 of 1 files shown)


musicinmybrain:
This seems to work for me.

(In Fedora, I’ll need to loosen the version bound to allow numpy 1.24.x until we get 1.25 packaged, but a distribution package doesn’t have the same version-compatibility concerns as a binary wheel on PyPI.)

PR #521, opened 2023-09-22
Remove Travis CI config

jarrodmillman:


effigies:
Should we remove appveyor as well?

yarikoptic:
FWIW, I don't see any coverage reporting here to see if removal of any CI drops it...  e.g. for https://github.com/datalad/datalad-container/pull/210 I had to spend some time to figure out all changes needed to ensure the same code coverage since we tested different code paths across CIs

effigies:
I don't think we can reasonably reconstruct what the old coverage was, since Travis has been down so long. We can set up codecov to at least track it moving forward.

effigies:
Actually, I guess we already had: https://app.codecov.io/gh/nipy/nipy/tree/master/nipy.

Let's see what we get...

PR #527, opened 2023-09-22
Port build system to Meson

stefanv:


meson.build
```diff
@@ -0,0 +1,47 @@
...
+cc = meson.get_compiler('c')
+
+# Check compiler is recent enough (see "Toolchain Roadmap" for details)
+if cc.get_id() == 'gcc'
+  if not cc.version().version_compare('>=8.0')
+    error('scikit-image requires GCC >= 8.0')
+  endif
+elif cc.get_id() == 'msvc'
+  if not cc.version().version_compare('>=19.20')
+    error('scikit-image requires at least vc142 (default with Visual Studio 2019) ' + \
```
jarrodmillman, line 23:
```suggestion
    error('nipy requires at least vc142 (default with Visual Studio 2019) ' + \
```

matthew-brett:
What say we merge this one?

Does it makes sense do delete all the `setup.py` files now?   Am I right in thinking that we have abandoned distutils build after this PR?

effigies:
How are you getting your metadata into the package? It looks like `setup.py` is still the only thing that could provide that.

matthew-brett:
I was assuming we should keep the root `setup.py`.

matthew-brett:
Ah - I'm running into build breakage when I remove the `setup.py` files, but I'm not sure why.   The output tree is essentially empty without the setup.py files - there are no subdirectories for the submodules.   I was surprised - why is that?

stefanv:
Build without setup.py looks OK to me; at least from the meson perspective.

stefanv:
@matthew-brett I raised the question around OpenBLAS usage from wheel on the NumPy community call this morning. That option is not viable quite yet, and the changes they are making for NumPy (which we can copy) will be available in Meson in a few months at earliest. But Ralf raised a relevant question, which is: why doesn't nipy use the scipy BLAS interface instead?

matthew-brett:
Ah - yes - the Scipy BLAS interface didn't exist at the time Alexis wrote this code.

matthew-brett:
But once you got OpenBLAS linking - did you get the errors I am seeing?

stefanv:
No, the build runs fine on my machine.

matthew-brett:
Yes, on mine too - but not on CI.

stefanv:
My latest commit should address the problem.

stefanv:
:rocket:

stefanv:
Note currently only links against OpenBLAS or lapack_lite, if that is missing. Can be improved to find Accelerate etc., along the lines of:

https://github.com/numpy/numpy/blob/main/numpy/meson.build#L94

stefanv:
Metadata, such as classifiers, removed in a4ea93c0884def364ca398bb2a9b50722dcbb514 will need to be put into `pyproject.toml`

jarrodmillman:
@matthew-brett Shoud we merge this and rebase the other PRs on main? It seems like this PR works. But maybe I am missing something.

jarrodmillman (approved):

matthew-brett:
@stefanv - thanks very much for doing this - it will save me a huge amount of time.    I am sorry to prolong the agony - but I'm still having difficulty building with the pytest changes - and I can't make any sense of the resulting error - that I also get locally.   Here's the current merge candidate build failure - do you have any ideas?  https://github.com/nipy/nipy/actions/runs/6408233434/job/17396753125

PR #528, opened 2023-09-23
Remove appveyor config

jarrodmillman:
I don't think we need it. If we want to keep using it, we should check that it is configured correctly:
https://ci.appveyor.com/project/nipy/nipy


effigies (approved):
No obvious value to me, with Windows runners on GitHub.

PR #538, opened 2023-10-04
MRG: merge of meson build into current main

matthew-brett:
Testing meson build with pytest changes


matthew-brett:
OK - all cleaned up and working.   The main problem was the deletion of the `fixes` and `external` modules, that was messing up Meson's build process.   I took the opportunity to delete the `.c` files derived from Cython that we were keeping around to allow build without Cython (it was like that in the old days), and add a page on debugging Meson builds, mainly for my reference.  We have also acquired a `conftest.py` file for `pytest` that needed adding to the Python sources in the `meson.build`.

stefanv:
@matthew-brett building via spin should print out the Meson commands used so you don't need to memorize them. And if there are any features needed in spin to help debug, let me know.

matthew-brett:
Aha - that sounds good - are there good Spin docs out there?

PR #540, opened 2023-10-15
Do not compare strings by identity (address)

musicinmybrain:
These comparisons against string literals may work in practice due to implementation details of string interning in CPython, but they are not reliable or correct, and they produce a `SyntaxWarning` in Python 3.12. Strings should always be compared by value (`==`/`!=`) rather than identity (`is`/`is not`).

### DIFF
--- a/nipy/labs/utils/reproducibility_measures.py
+++ b/nipy/labs/utils/reproducibility_measures.py
@@ -665,7 +665,7 @@ def group_reproducibility_metrics(
 
     from ..mask import intersect_masks
 
-    if ((len(variance_images) == 0) & (method != 'crfx')):
+    if len(variance_images) == 0 and method != 'crfx':
         raise ValueError('Variance images are necessary')
 
     nsubj = len(contrast_images)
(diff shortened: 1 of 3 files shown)


nipy/labs/utils/reproducibility_measures.py
```diff
@@ -665,7 +665,7 @@ def group_reproducibility_metrics(
 
     from ..mask import intersect_masks
 
-    if ((len(variance_images) == 0) & (method is not 'crfx')):
+    if ((len(variance_images) == 0) & (method != 'crfx')):
```
effigies, line 668: Might as well write Pythonically.

```suggestion
    if len(variance_images) == 0 and method != 'crfx':
```

effigies (approved):
LGTM. Small suggestion for while one of these lines is being touched anyway...

nipy/labs/utils/reproducibility_measures.py
musicinmybrain: You’re right, I was laser-focused on the particular fix and didn’t review the surrounding line.

I guess the even more Pythonic form would be:

```python
if not variance_images and method != 'crfx':
```

musicinmybrain: > ```python
> if not variance_images and method != 'crfx':
> ```

Which form should I change it to?

effigies: `len()` makes it clearer to a reader that a list or tuple or similar is expected, so I'd go with that.

musicinmybrain:
I’ll make a follow-up PR to fix other uses of bitwise operators as Boolean logical operators.

PR #541, opened 2023-10-15
Remove executable filesystem permissions from files that don’t need it

musicinmybrain:
The CLI entry points don’t have shebangs, so there is no way to execute them directly from the source tree (`./nipi/cli/foo.py`) without explicitly using the Python interpreter (`python ./nipy/cli/foo.py`). The executable bit in the filesystem permissions is therefore unnecessary.

It’s unlikely that making `coverage.yml` “executable” was intentional.


effigies (approved):
LGTM

PR #542, opened 2023-10-15
Style and pedantic correctness fixes in Boolean expressions

musicinmybrain:
Inspired by https://github.com/nipy/nipy/pull/540#pullrequestreview-1678802015.

I know it’s tedious to review changes that could affect control flow, but I do think these are worthwhile. I tried to be as careful as I could to watch out for values that could be tristate (`True`/`False`/`None`) and avoid changing the control flow for `None` values.


musicinmybrain:
I’ll review the test failures and amend the PR.

musicinmybrain:
I almost have this passing again, but the number of initial failures indicates that I didn’t audit these changes as carefully with respect to types and possible values as I needed to. I think I want to drop this PR for now. Please feel free to use it as a starting point if you like any of the changes it would have included.

PR #543, opened 2023-10-15
Fix backslash escaping issues in docstrings

musicinmybrain:
Escape backslashes that should be passed through to the documentation system rather than interpreted as backslash escape sequences in the Python string.

This fixes a number of `SyntaxWarning` instances about invalid escape sequences in Python 3.12.

An alternative would be to make these docstrings raw strings (`r"""foo"""`).

### DIFF
--- a/nipy/utils/perlpie.py
+++ b/nipy/utils/perlpie.py
@@ -1,6 +1,6 @@
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
-"""Perform a global search and replace on the current directory *recursively*.
+r"""Perform a global search and replace on the current directory *recursively*.
 
 This a small python wrapper around the `perl -p -i -e` functionality.
 I **strongly recommend** running `perlpie` on files under source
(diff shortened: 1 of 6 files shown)


DimitriPapadopoulos:
Alternatively add **r** for _raw string_ before **"**.

effigies:
My preference would also be for raw strings.

musicinmybrain:
I’m happy to convert these docstrings to raw strings when I have a little time in front of a computer.

nipy/utils/perlpie.py
```diff
@@ -71,15 +71,15 @@ def check_deps():
...
+    r"""Use perl to replace the oldstr with the newstr.
 
     Examples
     --------
 
     # To replace all occurences of 'import numpy as N' with 'import numpy as np'
     from nipy.utils import perlpie
-    perlpie.perl_dash_pie('import\\s+numpy\\s+as\\s+N', 'import numpy as np')
-    grind | xargs perl -pi -e 's/import\\s+numpy\\s+as\\s+N/import numpy as np/g'
+    perlpie.perl_dash_pie('import\s+numpy\s+as\s+N', 'import numpy as np')
```
effigies, line 81:
```suggestion
    perlpie.perl_dash_pie(r'import\s+numpy\s+as\s+N', 'import numpy as np')
```

effigies (approved):
LGTM. Small fix in one docstring.

PR #544, opened 2023-10-15
Cython: declare pointers one per line

musicinmybrain:
None of these declarations was incorrect, but Cython now warns about them. The developers are worried that people will write “`cdef int* a, b`” and think that “`b`” is an “`int*`” rather than an “`int`”. Reducing the number of warnings from Cython makes it easier to notice warnings that might represent real problems.

See also https://groups.google.com/g/cython-users/c/MA8I1wmoT0Q for discussion.

### DIFF
--- a/nipy/algorithms/statistics/_quantile.pyx
+++ b/nipy/algorithms/statistics/_quantile.pyx
@@ -48,7 +48,8 @@ def _quantile(X, double ratio, int interp=False, int axis=0):
     Y : array
       Array of quantiles
     """
-    cdef double *x, *y
+    cdef double *x
+    cdef double *y
     cdef long int size, stride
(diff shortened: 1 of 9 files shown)


effigies (approved):
I doubt there's much energy for actually modifying these files, so I think reducing build noise is probably worth 100+ lines of unnecessary duplication. If people do start working on this again, it would probably be a good idea to consider/adopt Dipy's [cython style guide](https://dipy.org/documentation/1.7.0/devel/coding_style_guideline/#cython-coding-style).

Will merge sooner or later if I hear no objections.

matthew-brett:
Thanks - I agree with Chris, just thought I'd merge as I was here.

PR #545, opened 2023-10-15
Improve const-correctness in C code

musicinmybrain:
The commit “Do not cast the result of PyArray_DATA()” is in this PR because it touches most of the same lines that needed to be touched to fix const-related warnings.

I’d like to suggest that a broader change to stop explicitly casting `void *` return values could make sense; I claim that they add clutter and repetition but don’t reduce errors. They are not required because `void *` happily converts implicitly to any data pointer type (in C, not C++). Such a change would affect at least `malloc()`, `calloc()`, `realloc()`, and `PyArray_ITER_DATA()`. However, I don’t want to get that tangled up with the const-correctness changes, especially if it’s not desired.


effigies (approved):
My C is pretty rusty, but this looks right to me.

PR #548, opened 2023-11-08
`is not` → `!=`

DimitriPapadopoulos:
A suggestion by pyupgrade that makes sense.


effigies:
Looks like no change.

PR #551, opened 2023-11-09
Remove setup.py information from info.py

matthew-brett:
Shift away from info.py as source of setup information.


pyproject.toml
```diff
@@ -1,20 +1,87 @@
...
 [build-system]
 build-backend = "mesonpy"
 requires = [
   "meson-python>=0.13",
+  "ninja",
   "setuptools",
   "cython>=3",
   # Newer than NEP29-minimum: compile against oldest numpy available
-  "numpy>=1.25; python_version > '3.8'",
+  "numpy>=1.22; python_version > '3.8'",
```
matthew-brett, line 80: @stefanv - am I missing something here?  1.22 is available for Python 3.9.

effigies: 1.25 supports compiling against older APIs.

pyproject.toml
```diff
@@ -1,20 +1,87 @@
...
+    'spin==0.3',
+    'build',
+]
+default = [
+    'numpy>=1.22',
+    'scipy>=1.8',
+    'nibabel>=3.2',
+    'sympy>=1.9',
+    'transforms3d'
+]
```
effigies, line 50: IMO build and default don't make much sense. `nipy[default]` is equivalent to `nipy` and now you have duplication.

I'm not sure what `nipy[build]` is adding, since to install it you will already need something capable of building, while also duplicating (and extending?) the `build-system.requires` table.

pyproject.toml
```diff
@@ -1,20 +1,87 @@
...
+    'pytest>=7.2',
+    'pytest-cov>=4.0',
+    'pytest-doctestplus'
+]
+
 [build-system]
 build-backend = "mesonpy"
 requires = [
   "meson-python>=0.13",
+  "ninja",
```
effigies, line 76: I'm able to build without this. Is `ninja` a dependency of `meson-python` that you just want to be explicit about? Do you actually depend on ninja-specifics, or could the other build tools stop using it without you needing to change anything?

pyproject.toml
```diff
@@ -1,20 +1,87 @@
...
+[[project.maintainers]]
+name = 'nipy developers'
+email = 'neuroimaging@python.org'
+
+[project.urls]
+homepage = 'https://nipy.org/nipy'
+documentation = 'http://nipy.org/nipy/documentation.html'
+source = 'https://github.com/nipy/nipy'
+download = 'https://pypi.org/project/nipy/#files'
+tracker = 'https://github.com/nipy/nipy/issues'
```
effigies, line 31: IIUC, these will be how they're actually displayed on PyPI. Capitals would look nicer, I think. And you can use `"Several Words" = 'https://...'` if you want to include spaces.

```suggestion
Homepage = 'https://nipy.org/nipy'
Documentation = 'http://nipy.org/nipy/documentation.html'
Source = 'https://github.com/nipy/nipy'
Download = 'https://pypi.org/project/nipy/#files'
Tracker = 'https://github.com/nipy/nipy/issues'
```

effigies (commented):
Some quick notes. Nothing blocking.

pyproject.toml
matthew-brett: @stefanv - any thoughts?  I copied this brainlessly from scikit-image: https://github.com/scikit-image/scikit-image/blob/main/pyproject.toml#L46

pyproject.toml
matthew-brett: I was getting errors trying to build without it, for the wheel builds.   I'm afraid I don't understand the build mechanics well enough to respond.   If ninja is missing, with meson-python use something else?

effigies: How are you building? I'm using `python -m build`.

effigies: Hmm. Possibly an old pip? I just created a fresh environment:

```
$ mamba create -n tmp python=3.11
$ mamba activate tmp
$ pip wheel .
Processing /home/chris/Projects/nipy/nipy
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: nipy
  Building wheel for nipy (pyproject.toml) ... done
  Created wheel for nipy: filename=nipy-0.5.1.dev1-cp311-cp311-linux_x86_64.whl size=1916097 sha256=7ff5b2a9a9fe053c913dd2e487ee8c8fd11991893cf96e0f806846c79d17cd8b
  Stored in directory: /tmp/pip-ephem-wheel-cache-sn4mrur1/wheels/a1/58/f7/1af5aca9b8327defaf19028fc61c6f4850bd54b33e92f89a43
Successfully built nipy
$ pip --version
pip 23.3.1 from /home/chris/mambaforge/envs/tmp/lib/python3.11/site-packages/pip (python 3.11)
```

stefanv:
I'm curious why dropping ninja works. You need to get it into your environment somehow, and it is not installed as a dependency of meson or meson-python.

stefanv:
I just verified that, if ninja is not presence, it will fail on a build. It can be confusing to test, because if nox is already on the PATH (from a different venv), it will still be picked up and used.

To see that in action, clone https://github.com/stefanv/test-ninja-missing-compile-failure and run `nox`, ensuring that ninja is NOT on the PATH.

effigies:
I don't have ninja installed in my path, and using `pipx run build` to run in a completely clean environment, I see:

```console
$ pipx run build
[...]
Build targets in project: 16

nipy 0.5.1.dev1

  User defined options
    Native files: /tmp/build-via-sdist-48bra0w8/nipy-0.5.1.dev1/.mesonpy-x0vcp63g/meson-python-native-file.ini
    buildtype   : release
    b_ndebug    : if-release
    b_vscrt     : md

Found ninja-1.11.1.git.kitware.jobserver-1 at /tmp/build-env-bge1hi3c/bin/ninja
```

So it does seem to be getting pulled in somewhere. I don't know how right now...

effigies:
Hmm. Missed an important bit:

```console
$ pipx run build 
* Creating venv isolated environment...
* Installing packages in isolated environment... (cython>=3, meson-python>=0.13, numpy==1.22; python_version >= '3.7' and python_version < '3.9', numpy>=1.25; python_version > '3.8', setuptools)
* Getting build dependencies for sdist...
* Installing packages in isolated environment... (ninja >= 1.8.2)
```

I would guess meson is injecting it.

effigies:
> To see that in action, clone https://github.com/stefanv/test-ninja-missing-compile-failure and run `nox`, ensuring that ninja is NOT on the PATH.

I can confirm that `pipx run nox` doesn't work, but `pipx run build` does.

matthew-brett:
Sounds like it would be safer to add it back, no?

effigies:
Sure, why not.

stefanv:
OK, so it looks like you get ninja pulled in when building via `build`. On scikit-image, we generate `requirement.txt` files from our `pyproject.toml` to make it easy to install build dependencies, so we need ninja there for when building the project via `spin`.

pyproject.toml
matthew-brett: @stefanv - any reason not to just delete the `build` and `default` sections of the `optional-dependencies`?

stefanv:
No, I think skimage may have done the same. We used to use it to generate requirements, but those are now picked up from the correct locations in the file.

matthew-brett:
Thanks Stefan.   Chris - does this look right to you now?

effigies:
Yup, that all looks fine to me.

PR #552, opened 2023-11-10
ENH: Replace saving mayavi figure to a temporary file with mlab.screenshot()

fazledyn-or:
## Details
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [maps_3d.py](https://github.com/nipy/nipy/blob/main/nipy/labs/viz_tools/maps_3d.py#L120), there is a method that creates a temporary file using an unsafe API `mktemp`. The use of this method is discouraged in the [Python documentation](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). iCR suggested that a temporary file should be created using `mkstemp` which is a [safe API](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp). iCR replaced the usage of mktemp with `mkstemp`.

#### Resources Related to `mktemp`
- [POC - Improper Method Call - python - mktemp.mp4](https://drive.google.com/file/d/16YKuKgv7iItWcac2biRyi2Cm8upJ3HP_/view?usp=share_link)
- [MISC - Python - Taking a peek under the tempfile library.mp4](https://drive.google.com/file/d/1ns2_x_ZDvzFwj0eNNxyay-HUMclLasNT/view?usp=share_link)

## Changes
- Replaced `mktemp()` method with `mkstemp()`

## Previously Found & Fixed
- https://www.github.com/invesalius/invesalius3/pull/679
- https://www.github.com/Azure/azure-linux-extensions/pull/1816
- https://www.github.com/celery/billiard/pull/394

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.


effigies:
This seems correct, although mayavi now has its own `screenshot()` function (and has since 2010), so the better solution is probably to use that.

fazledyn-or:
Understandable.

Well, if anyone finds this method and wants to implement a `screenshot` method for themselves, chances are they are going to use `mktemp`. In that case, by proposing the `mkstemp` method, we are indirectly helping them avoid using deprecated methods.

However, the final decision regarding this PR is up to you.

Thanks!

nipy/labs/viz_tools/maps_3d.py
```diff
@@ -117,14 +117,15 @@ def m2screenshot(mayavi_fig=None, mpl_axes=None, autocrop=True):
     if mpl_axes is not None:
         plt.axes(mpl_axes)
 
-    filename = tempfile.mktemp('.png')
+    fd, filename = tempfile.mkstemp('.png')
```
matthew-brett, line 120: Will this also work on Windows?   I mean - writing to the `filename` while the file is already open with file descriptor `fd`?   And then reading?

nipy/labs/viz_tools/maps_3d.py
```diff
@@ -117,14 +117,15 @@ def m2screenshot(mayavi_fig=None, mpl_axes=None, autocrop=True):
...
     mlab.savefig(filename, figure=mayavi_fig)
     image3d = plt.imread(filename)
     if autocrop:
         bg_color = mayavi_fig.scene.background
         image3d = autocrop_img(image3d, bg_color)
     plt.imshow(image3d)
     plt.axis('off')
+    os.close(fd)
     os.unlink(filename)
```
effigies, line 129: Sorry, this is what I was proposing:

```suggestion
    image3d = mlab.screenshot()
    if autocrop:
        bg_color = mayavi_fig.scene.background
        image3d = autocrop_img(image3d, bg_color)
    plt.imshow(image3d)
    plt.axis('off')
```

nipy/labs/viz_tools/maps_3d.py
fazledyn-or: It also works on Windows. Opening/writing to a file that already has a file-descriptor opened is possible.

You can have a look at this example - 
image

Here, I'm opening the same file twice. In the context manager, when I open the file (`f`), it gets the file-descriptor value of 4 assigned to it. And I while writing to the file, the first `fd: 3` is still opened. So, it's not a problem.

nipy/labs/viz_tools/maps_3d.py
fazledyn-or: Alright, let me add these changes. It looks much cleaner.

matthew-brett:
I hate to ask - but has anyone tested this change to see if it works?

effigies:
Tested with:

```
python -c 'from nipy.labs.viz_tools.activation_maps import demo_plot_map; demo_plot_map(do3d=True)'
```

Took some fixes.

matthew-brett:
Ouch - and thank you very much for testing that.  I guess we could add that as a smoketest provisional on mayavi being present.   But given these will go away soon, let's just merge with the improvment.

PR #558, opened 2023-12-23
MNT: Test and advertise support for Python 3.8+

### LINKED ISSUE #557: Supported Python versions?
I understand nipy supports Python ≥ 3.7:
https://github.com/nipy/nipy/blob/ca70df937b6943d2fbc37b65c463bb78bab28f22/pyproject.toml#L62-L67

However, this is not clearly documented.

Worse, the ruff setting [target-version](https://docs.astral.sh/ruff/settings/#target-version), the minimum Python version to target, is set to Python 3.9:
https://github.com/nipy/nipy/blob/ca70df937b6943d2fbc37b65c463bb78bab28f22/pyproject.toml#L77-L79

How about adding the [`requires-python`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) key in the [`[project]`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata) table? As far as I can see, [mesonpy understands it](https://mesonbuild.com/meson-python/tutorials/introduction.html#configuring-our-python-package).

effigies:
Nipy builds and tests clean on Python 3.8, so this just acknowledges and advertises that. We could also follow [SPEC0](https://scientific-python.org/specs/spec-0000/), which would have us drop Python <3.10 at this point. IMO Nipy doesn't need to be as conservative as nibabel here, so bumping to 3.10+ would be fine.

Fix picked up by `ruff` included.

Closes #557.


pyproject.toml
```diff
@@ -63,8 +68,8 @@ requires = [
   # version.  See:
   # https://numpy.org/doc/stable/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default
   "numpy>=1.25; python_version > '3.8'",
-  # NEP29-minimum as of Sep 21, 2023
-  "numpy==1.22; python_version >= '3.7' and python_version < '3.9'",
+  # SPEC0-minimum as of Dec 23, 2023
+  "numpy==1.22; python_version <= '3.8'",
```
matthew-brett, line 72: Shouldn't this be `python_version >= '3.8'"`?

matthew-brett (commented):
Point of clarification only.

pyproject.toml
effigies: No. Python 3.9 is supported by numpy 1.25, so it's covered by the previous line. I could set this `== '3.8'`, but I'm not 100% sure what the result would be if someone tried to build with Python 3.7, as it would think it could build without numpy at all. OTOH, it's possible that the `python-requires` would prevent from getting that far.

matthew-brett: Ok - yes - sorry - I remember.

pyproject.toml
```diff
@@ -76,7 +81,7 @@ nipy_diagnose = 'nipy.cli.diagnose:main'
 
 [tool.ruff]
 line-length = 88
-target-version = 'py39'
+target-version = 'py38'
```
DimitriPapadopoulos, line 84: Isn't ruff able to catch `python-requires = ">=3.8"`? I'll investigate that.

effigies: Yes, it is. I forgot.

PR #566, opened 2024-07-25
RF: nibabel.onetime.auto_attr -> functools.cached_property

effigies:
Nipy has a minimum Python of 3.8, so we can use the stdlib `functools.cached_property`.


matthew-brett:
Nice - thanks.

PR #565, opened 2024-06-18
Numpy 20 fixes

matthew-brett:
Working on build / compile and test for Numpy 2.0.

PR #568, opened 2024-09-04
fffpy.c: fix build failure with gcc 14.

emollier:
As initially reported in [Debian bug #1075317], nipy fails to build from source with the following error since introduction of gcc 14, and the promotion of several warnings as errors:

	../lib/fff_python_wrapper/fffpy.c:17:3: error: returning ‘void *’ from a function with return type ‘IMP_OUT’ {aka ‘int’} makes integer from pointer without a cast [-Wint-conversion]
	   17 |   import_array();
	      |   ^~~~~~~~~~~~

This looks to be related to the logic revolving around setting the type of IMP_OUT, as it does not know about the present import_array macro "return type" in contemporary Numpy versions.  Since Python 2 is obsolete, this patch removes the Python major selector entirely to replace by a hard wiring of the IMP_OUT type.

This may not be 100% suitable to everyone, but the time of introduction of the void* return type to the import_array macro in Numpy is unclear.

[Debian bug #1075317]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1075317

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           allow-prereleases: true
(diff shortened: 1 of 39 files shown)


effigies:
Included in #565.

PR #569, opened 2024-09-04
adjust coordinate_map.py and tests with sympy 1.13

emollier:
Since version 1.13, SymPy makes a hard distinction between integers and floating point numbers.  This causes a couple of tests to fail, with the most relevant part of the error showing:

	self = AffineTransform(
	   function_domain=CoordinateSystem(coord_names=('x', 'y', 'z'), name='', coord_dtype=object),
	   fun...ne=array([[3, 0, 0, 0],
	                 [0, 4, 0, 0],
	                 [0, 0, 5, 0],
	                 [0, 0, 0, 1]])
	)
	function_domain = CoordinateSystem(coord_names=('x', 'y', 'z'), name='', coord_dtype=object)
	function_range = CoordinateSystem(coord_names=('i', 'j', 'k'), name='', coord_dtype=object)
	affine = array([[1.00000000000000, 0, 0, 0],
	       [0, 0, 1.00000000000000, 0],
	       [0, 1.00000000000000, 0, 0],
	       [0, 0, 0, 1.00000000000000]], dtype=object)

	    def __init__(self, function_domain, function_range, affine):
	        """ Initialize AffineTransform
	[…]
	        if not np.all(affine[-1] == bottom_row):
	>           raise ValueError('the homogeneous transform should have bottom=' + \
	                             f'row {repr(bottom_row)}')
	E           ValueError: the homogeneous transform should have bottom=row array([0, 0, 0, 1])

This change implements an adjustments to avoid the AffineTransform constructor to trip on the mismatch; as a bonus, the error message is extended to be a little bit more helpful.  Following up on this change and sympy's, a couple test outputs needed to be normalized.

### DIFF
--- a/nipy/core/reference/coordinate_map.py
+++ b/nipy/core/reference/coordinate_map.py
@@ -558,10 +558,11 @@ class AffineTransform:
                              'affine matrix shape')
         # Test that it is actually an affine mapping in homogeneous
         # form
-        bottom_row = np.array([0]*self.ndims[0] + [1])
+        bottom_row = np.array([0]*self.ndims[0] + [1.], dtype=dtype)
         if not np.all(affine[-1] == bottom_row):
             raise ValueError('the homogeneous transform should have bottom=' + \
-                             f'row {repr(bottom_row)}')
+                             f'row {repr(bottom_row)}' + ' but it has bottom=' \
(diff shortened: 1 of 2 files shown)


effigies:
I think the meat of this one was covered by #565 as well.

PR #574, opened 2024-09-21
Apply ruff rules (RUF)

DimitriPapadopoulos:


nipy/algorithms/statistics/tests/test_intrinsic_volumes.py
```diff
@@ -190,7 +192,7 @@ def test_ec():
         assert_almost_equal(f(box1), 1)
         # While we're here, test we can use different dtypes, and that values
         # other than 0 or 1 raise an error.
-        for dtt in sum([np.sctypes[t] for t in ('int', 'uint', 'float')], []):
+        for dtt in functools.reduce(operator.iadd, [np.sctypes[t] for t in ('int', 'uint', 'float')], []):
```
effigies, line 195: This is definitely not more readable. I'd be okay to switching to `chain.from_iterable` from `itertools`, though.

effigies (commented):
Needs rebase or rerun. I disagree with `reduce(iadd, ..., [])` instead of `sum(..., [])`

nipy/algorithms/statistics/tests/test_intrinsic_volumes.py
DimitriPapadopoulos: Actually, [quadratic-list-summation (RUF017)](https://docs.astral.sh/ruff/rules/quadratic-list-summation/) is not about readability:
> The use of `sum()` to flatten lists of lists is quadratic in the number of lists, as `sum()` creates a new list for each element in the summation.
> 
> Instead, consider using another method of flattening lists to avoid quadratic complexity. The following methods are all linear in the number of lists:
> * `functools.reduce(operator.iadd, lists, [])`
> * `list(itertools.chain.from_iterable(lists))`
> * `[item for sublist in lists for item in sublist]`

But then that's hardly an issue for these small lists.

PR #584, opened 2024-10-04
Test on Python 3.13

DimitriPapadopoulos:
Python 3.13 is not available yet in some runners, use Python 3.12 instead.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.12"]
+        python-version: ["3.13"]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.13"]
+        python-version: ["3"]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.12"]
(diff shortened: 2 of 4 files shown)


matthew-brett:
It looks like Python 3.13 is not yet available on Ubuntu. 

```
Installed versions
  Version 3.13 was not found in the local cache
  Error: The version '3.13' with architecture 'x64' was not found for Ubuntu 22.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

 Let's stick to Python 3.12 as the default version to test on, for now, but test on Python 3.13 for the platforms on which it is currently available.   How are we for wheels on Python 3.13?

DimitriPapadopoulos:
Python 3.13 will be released on 7 October. I suspect it will be available for all tests soon after that.

.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   test:
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
effigies, line 26:
```suggestion
        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev"]
```

.github/workflows/doc-build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.12"]
+        python-version: ["3.13"]
```
effigies, line 18:
```suggestion
        python-version: ["3"]
```

.github/workflows/coverage.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.12"]
+        python-version: ["3.13"]
```
effigies, line 18:
```suggestion
        python-version: ["3"]
```

effigies:
Oh I misunderstood and thought 3.13 wasn't installing.

.github/workflows/test.yml
matthew-brett: Yes - this does not appear to be necessary ...

matthew-brett:
Thanks @DimitriPapadopoulos

